PR #2, opened 2016-02-03
turn on cached resolution in SBT.  See:

nh13:
- http://www.scala-sbt.org/0.13/docs/Cached-Resolution.html

[Review on Reviewable](https://reviewable.io/reviews/fulcrumgenomics/dagr/2)


nh13:
This should speed up our build process in both development and CI.

PR #10, opened 2016-02-03
Changed the CLP pipelines to provide the output directory to the pare…

tfenne:
…nt constructor.

### DIFF
--- a/pipelines/src/main/scala/dagr/pipelines/CreateUnmappedBamFromFastqPipeline.scala
+++ b/pipelines/src/main/scala/dagr/pipelines/CreateUnmappedBamFromFastqPipeline.scala
@@ -51,9 +51,10 @@ class CreateUnmappedBamFromFastqPipeline(
   @Arg(flag="l", doc="The name of the library.")                                  val library: String,
   @Arg(flag="p", doc="The platform unit (@RG.PU).")                               val platformUnit: List[String],
   @Arg(doc="Path to a temporary directory.")                                      val tmp: Path,
-  @Arg(flag="o", doc="The output prefix for files that are kept.")                val output: Path,
-  @Arg(doc="Path to the unmapped BAM.  Use the output prefix if none is given. ") var unmappedBam: Option[PathToBam] = None
-) extends Pipeline {
+  @Arg(flag="o", doc="The output directory in which to write files.")             val output: DirPath,
+  @Arg(doc="The filename prefix for output files. Library is used if omitted.")   val basename: Option[FilenamePrefix] = None,
+  @Arg(doc="Path to the unmapped BAM. Use the output prefix if none is given. ")  var unmappedBam: Option[PathToBam] = None
+) extends Pipeline(Some(output)) {
 
   name = CreateUnmappedBamFromFastqPipeline.NAME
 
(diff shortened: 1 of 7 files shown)


pipelines/src/main/scala/dagr/pipelines/CreateUnmappedBamFromFastqPipeline.scala
```diff
@@ -53,7 +53,7 @@ class CreateUnmappedBamFromFastqPipeline(
   @Arg(doc="Path to a temporary directory.")                                      val tmp: Path,
   @Arg(flag="o", doc="The output prefix for files that are kept.")                val output: Path,
   @Arg(doc="Path to the unmapped BAM.  Use the output prefix if none is given. ") var unmappedBam: Option[PathToBam] = None
-) extends Pipeline {
+) extends Pipeline(Some(output.toAbsolutePath.getParent)) {
```
nh13: Perhaps we should make an implicit method on a `Path`: `getAbsoluteParentPath`?  Your call.

pipelines/src/main/scala/dagr/pipelines/CreateUnmappedBamFromFastqPipeline.scala
```diff
@@ -53,7 +53,7 @@ class CreateUnmappedBamFromFastqPipeline(
   @Arg(doc="Path to a temporary directory.")                                      val tmp: Path,
   @Arg(flag="o", doc="The output prefix for files that are kept.")                val output: Path,
```
nh13: I am not satisfied with the dual use of `output`.  Perhaps we use `prefix` and `output`, with the former a `Path` prefix and the latter a directory.

pipelines/src/main/scala/dagr/pipelines/DNAResequencingFromUnmappedBamPipeline.scala
```diff
@@ -56,7 +56,7 @@ class DnaResequencingFromUnmappedBamPipeline(
   @Arg(flag="t", doc="Target intervals to run HsMetrics over.")    val targetIntervals: Option[PathToIntervals],
   @Arg(doc="Path to a temporary directory.")                       val tmp: Path,
   @Arg(flag="o", doc="The output prefix for files that are kept.") val output: Path
```
nh13: Can you change `Path` to `DirPath`?

pipelines/src/main/scala/dagr/pipelines/CreateUnmappedBamFromFastqPipeline.scala
nh13: Alternatively, we could have a type called `PrefixPath`?

nh13:
@tfenne back to you, make the changes you feel you want, and then merge it.

pipelines/src/main/scala/dagr/pipelines/CreateUnmappedBamFromFastqPipeline.scala
tfenne: I would be more tempted to add a method in `Io` like:

``` scala:
def pathTo(path: String) : Path = Paths.get(path).toAbsolutePath
```

and use that everywhere we can instead of Paths.get.  What do you think?

PR #13, opened 2016-02-03
Fixed a problem where Io.assertCanWriteFile failed on relative paths.

tfenne:
I wanted to write a test for this, but since it involves relative paths, it's rather difficult.  I tested the fix in the sbt console instead, and it works.


nh13:
:+1:

PR #15, opened 2016-02-04
downgrading coveralls:

nh13:
https://github.com/scoverage/sbt-coveralls/issues/73


nh13:
Trying this again, will revert if it fails.

PR #24, opened 2016-02-04
Allow pipeline packages to be specific via config.

tfenne:


core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -33,8 +33,10 @@ import dagr.core.execsystem.{Cores, Memory}
 import dagr.core.util.{Io, LazyLogging}
 
 import scala.collection.SortedSet
+import scala.collection.JavaConversions._
 import scala.reflect.runtime.universe._
 
+
```
nh13, line 39: !

nh13:
:+1:

PR #30, opened 2016-02-04
Tidied up @CLP parameters.

tfenne:
Fixed the fact that @CLPConstructor was a scala annotation(!).
Changed CommandLineTaskGroup to PipelineGroup and renamed some stuff.
Changed omitFromCommandLine to hidden everywhere.


nh13:
I think `sbt doc` will fail, which likely will hold up a release.  Bummer, but I could certainly have help figuring out a work around.  Perhaps add `sbt doc` to the `.travis.yml` so it errors when generating doc would fail?

nh13:
Also, look at the output help message, which looks horrible due to how we had to mangle the docs to get `sbt doc` to work.

pipelines/src/main/scala/dagr/pipelines/GermlineVariantCallingPipeline.scala
```diff
@@ -38,11 +38,10 @@ import dagr.tasks.vc.{FilterFreeBayesCalls, FreeBayesGermline}
...
 @CLP(
-  summary = """Calls germline variants in a single sample and then hard filters them.
-For GATK, calls are made using the GATK's HaplotypeCaller and filtering with
-Picard's FilterVcf.  For FreeBayes, calls are made using bcbio's best practices.
-Optionally, variants are assessed with Picard's GenotypeConcordance tool.""",
-  oneLineSummary = "Calls and filters germline variants."
+  description = "Calls germline variants in a single sample and then hard filters them. " +
+                   "For GATK, calls are made using the GATK's HaplotypeCaller and filtering with " +
+                   "Picard's FilterVcf.  For FreeBayes, calls are made using bcbio's best practices. " +
+                   "Optionally, variants are assessed with Picard's GenotypeConcordance tool."
```
nh13: scaladoc wont like this.

nh13:
Sorry to put the burden of `sbt doc` onto this PR, so lets figure out a way forward.  Back to you @tfenne .

tfenne:
Stupid scaladoc, this is really lame.  Probably the easiest option would be to exclude the example workflows by package from the scaladoc generation until such time as this works.  I don't think that would be so bad.  Thoughts?

tfenne:
I've changed that one pipeline back to using a multiline string (formatted in a way that offends me slightly less), and verified that `sbt doc` now runs successfully.  Definitely need a better long term solution though.

PR #33, opened 2016-02-05
adding tests in sopt and core

nh13:


sopt/src/main/scala/dagr/sopt/ArgTokenCollator.scala
```diff
@@ -71,7 +71,6 @@ class ArgTokenCollator(argTokenizer: ArgTokenizer) extends Iterator[Try[ArgOptio
       case Success(ArgOptionAndValue(name, value)) => values += value; Success(name)
       case Success(ArgValue(value)) => Failure(new OptionNameException(s"Illegal option: '$value'"))
       case Failure(ex) => Failure(ex)
-      case Success(tok) => throw new IllegalStateException(s"Unknown token: '${tok.getClass.getSimpleName}")
```
nh13, line 74: can't happen since `ArtTokenizer.Token`  is a sealed class.

nh13:
@tfenne I am merging this as this is really not that exciting of a review, given its 90% tests.

core/src/main/scala/dagr/core/execsystem/ResourceSet.scala
```diff
@@ -24,7 +24,7 @@
 package dagr.core.execsystem
 
 object ResourceSet {
-  def apply(that: ResourceSet): ResourceSet = new ResourceSet(cores = that.cores, memory = that.memory)
+  def apply(that: ResourceSet): ResourceSet = new ResourceSet(cores = Cores(that.cores.cores), memory = Memory(that.memory.value))
```
tfenne, line 27: Given that ResourceSet is immutable, perhaps it just makes sense to eliminate this apply method entirely?

core/src/main/scala/dagr/core/cmdline/parsing/ParsingUtil.scala
```diff
@@ -35,15 +36,17 @@ import scala.collection.immutable.Map
...
 
 /** Variables and Methods to support command line parsing */
-private[parsing] object ParsingUtil {
+private[parsing] object ParsingUtil extends Configuration  {
   /** Gets the [[CLPAnnotation]] annotation on this class */
   def findClpAnnotation(clazz: Class[_]): Option[CLPAnnotation] = {
     ReflectionUtil.findJavaAnnotation(clazz, classOf[CLPAnnotation])
   }
 
+  private lazy val printColor: Boolean = optionallyConfigure[Boolean](Configuration.Keys.ColorStatus).getOrElse(false)
```
tfenne, line 45: Two things here:
1) Perhaps we should just add this to the reference.conf and not use a default here?
2) If you want to keep the default in code there's also `configuration[Boolean](Configuration.Keys.ColorStatus, false)`

tfenne:
Thanks for doing all this.  I read through anyway just to see what you were adding.  Made a couple of small suggestions you can chose to ignore if you like.

PR #48, opened 2016-02-05
modifying access levels for various methods, classes, objects (etc.) in

nh13:
sopt and core

[Review on Reviewable](https://reviewable.io/reviews/fulcrumgenomics/dagr/48)

### DIFF
--- a/core/src/main/scala/dagr/core/cmdline/CommandLineException.scala
+++ b/core/src/main/scala/dagr/core/cmdline/CommandLineException.scala
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2015 Fulcrum Genomics LLC
+ * Copyright (c) 2015-6 Fulcrum Genomics LLC
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -23,63 +23,63 @@
  */
 package dagr.core.cmdline
 
-object CommandLineException {
-  private[cmdline] def format(msg: String, e: Exception): String = {
+private[core] object CommandLineException {
+  private[core] def format(msg: String, e: Exception): String = {
     msg + ": " + e.getMessage + "\n" + e.getStackTrace.mkString("\n")
   }
 }
 
 /** Base class for all exceptions thrown by the command line parsing */
-class CommandLineException(msg: String) extends RuntimeException(msg) {
+private[core] class CommandLineException(msg: String) extends RuntimeException(msg) {
   def this(msg: String, e: Exception) {
     this(CommandLineException.format(msg, e))
   }
 }
 
 /** Exception thrown when the user forgets to specify an argument */
-case class MissingArgumentException(msg: String) extends CommandLineException(msg = msg) {
+private[core] case class MissingArgumentException(msg: String) extends CommandLineException(msg = msg) {
   def this(msg: String, e: Exception) {
     this(CommandLineException.format(msg, e))
   }
 }
 
 /** Exception thrown when the user forgets to specify an annotation */
-case class MissingAnnotationException(msg: String) extends CommandLineException(msg = msg) {
+private[core] case class MissingAnnotationException(msg: String) extends CommandLineException(msg = msg) {
   def this(msg: String, e: Exception) {
     this(CommandLineException.format(msg, e))
   }
 }
 
 /** Exception thrown when trying to convert to specific value */
-case class ValueConversionException(msg: String) extends CommandLineException(msg = msg) {
+private[core] case class ValueConversionException(msg: String) extends CommandLineException(msg = msg) {
   def this(msg: String, e: Exception) {
     this(CommandLineException.format(msg, e))
   }
 }
 
 /** Exception thrown when something internally goes wrong with command line parsing */
-case class CommandLineParserInternalException(msg: String) extends CommandLineException(msg = msg) {
+private[core] case class CommandLineParserInternalException(msg: String) extends CommandLineException(msg = msg) {
   def this(msg: String, e: Exception) {
     this(CommandLineException.format(msg, e))
   }
 }
 
 /** Exception thrown when something there is user error (never happens) */
-case class UserException(msg: String) extends CommandLineException(msg = msg) {
+private[core] case class UserException(msg: String) extends CommandLineException(msg = msg) {
   def this(msg: String, e: Exception) {
     this(CommandLineException.format(msg, e))
   }
 }
 
 /** Exception thrown when something the user gives a bad value */
-case class BadArgumentValue(msg: String) extends CommandLineException(msg = msg) {
+private[core] case class BadArgumentValue(msg: String) extends CommandLineException(msg = msg) {
   def this(msg: String, e: Exception) {
     this(CommandLineException.format(msg, e))
   }
 }
 
 /** Exception thrown when something the annotation on a field is incorrect. */
-case class BadAnnotationException(msg: String) extends CommandLineException(msg = msg) {
+private[core] case class BadAnnotationException(msg: String) extends CommandLineException(msg = msg) {
   def this(msg: String, e: Exception) {
     this(CommandLineException.format(msg, e))
   }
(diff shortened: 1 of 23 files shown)


core/src/main/scala/dagr/core/cmdline/CommandLineException.scala
```diff
@@ -23,63 +23,63 @@
  */
 package dagr.core.cmdline
 
-object CommandLineException {
-  private[cmdline] def format(msg: String, e: Exception): String = {
+private[core] object CommandLineException {
```
tfenne, line 26: If the object is `private[core]` do we gain anything by making each exception `private[core]`?  Since no-one can import the object's name or children?  Just thinking that reducing the repetition will make life easier if we change our minds later.

core/src/main/scala/dagr/core/cmdline/parsing/CommandLineParser.scala
```diff
@@ -62,9 +58,9 @@ object ParseResult extends Enumeration {
   val Version = Value // A request to display version was encountered
 }
 
-object CommandLineParser {
-  private[parsing] def parseArgumentsFile(lines: Traversable[String]): Traversable[String] = {
-    val args: ListBuffer[String] = new ListBuffer[String]
+private[parsing] object CommandLineParser {
+  def parseArgumentsFile(lines: Traversable[String]): Traversable[String] = {
+    val args: mutable.ListBuffer[String] = new mutable.ListBuffer[String]
```
tfenne, line 63: I'm just curious why this change?  There's no immutable ListBuffer so it was non-ambiguous before.

core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -40,7 +40,9 @@ import scala.reflect.runtime.universe.{TypeTag, typeOf}
   * Companion object to the Configuration trait that keeps track of all configuration keys
   * that have been requested so that they can be reported later if desired.
   */
-private[core] object Configuration extends ConfigurationLike {
+object Configuration extends ConfigurationLike {
+  // Developer Note: [[Configuration]] is not private so that [[Configuration.Keys]] is public
```
tfenne, line 45: I'm not convinced of this - the point of having this private was to prevent pipeline developers from using it instead of mixing in `Configuration`.  Can we leave it `private[core]` or perhaps even `private[dagr]`?

core/src/main/scala/dagr/core/execsystem/TaskRunner.scala
```diff
@@ -31,8 +31,73 @@ import dagr.core.util.LazyLogging
 
 import scala.collection.mutable
 
+private object TaskRunner {
+
```
tfenne: \n

core/src/main/scala/dagr/core/util/PathUtil.scala
```diff
@@ -40,8 +40,8 @@ object PathUtil {
     * @return the filename without illegal characters
     */
   def sanitizeFileName(fileName: String, illegalCharacters: String = PathUtil.illegalCharacters, replacement: Option[Char] = Some('_')): String = replacement match {
-    case None     => fileName.filter(c => !illegalCharacters.contains(c))
-    case Some(r)  => fileName.map(c => if (illegalCharacters.contains(c)) r else c)
+    case None => fileName.filter(c => !illegalCharacters.contains(c))
```
tfenne: IntelliJ strikes again?

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
```diff
@@ -26,7 +26,7 @@ class ReflectiveBuilder[T](val clazz: Class[T]) {
   protected val typeSignature = clazzSymbol.typeSignature
   protected val constructor   = findConstructor
   protected val jConstructor  = toJavaConstructor(constructor)
-  lazy val argumentLookup = new ArgumentLookup[ArgumentDescription]()
+  private[util] lazy val argumentLookup = new ArgumentLookup[ArgumentDescription]()
```
tfenne: I think this should remain at the same access level at the class - i.e. public.  Without it, the class is effectively useless.

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
```diff
@@ -145,7 +145,7 @@ class ReflectiveBuilder[T](val clazz: Class[T]) {
   * Class that encapsulates the information about the arguments to a CommandLineTask and provides
   * various ways to access the data.
   */
-class ArgumentLookup[ArgType <: Argument](args: ArgType*) {
+private[core] class ArgumentLookup[ArgType <: Argument](args: ArgType*) {
```
tfenne: Same here.  Either the whole file goes private, or stays public.

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
```diff
@@ -187,7 +187,7 @@ class ArgumentLookup[ArgType <: Argument](args: ArgType*) {
   * @param typeName the String name that identifies the unit type. May be the simple name of a class, or if a type
   *                 alias is used, the string name of the type alias
   */
-class Argument(val declaringClass: Class[_],
+private[core] class Argument(val declaringClass: Class[_],
```
tfenne: And here.

tfenne:
Looks good overall - thank you for doing this.  The only thing I care strongly about is the `ReflectiveBuilder` changes.  IMHO either the changes there should get reverted _or_ we should decide that the whole thing is private to dagr and make it so.

nh13:
`ReflectiveBuilder` is an interesting case.  It keys on the annotations that we have defined (ex. `ArgAnnotation`), but that annotation is not used anywhere (so I have removed it in the latest commit), which now makes this class standalone and I now agree it should be public.

---

Review status: 0 of 23 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

_[core/src/main/scala/dagr/core/cmdline/CommandLineException.scala, line 26 [r1]](https://reviewable.io:443/reviews/fulcrumgenomics/dagr/48#-K9n4rmG6Q7DTfFfATh9-r1-26:-K9sEM7U0QkyzfE90RIY:2016152244) ([raw file](https://github.com/fulcrumgenomics/dagr/blob/c29b2cb93667b7b5157f57599422b6aa1383f129/core/src/main/scala/dagr/core/cmdline/CommandLineException.scala#L26)):_
The exceptions are defined outside of the object, so they don't need to import the object to get at these exceptions.

---

_[core/src/main/scala/dagr/core/cmdline/parsing/CommandLineParser.scala, line 63 [r1]](https://reviewable.io:443/reviews/fulcrumgenomics/dagr/48#-K9n4rmG6Q7DTfFfAThB-r1-63:-K9sETU13U_0-bJRhYok:-27765462) ([raw file](https://github.com/fulcrumgenomics/dagr/blob/c29b2cb93667b7b5157f57599422b6aa1383f129/core/src/main/scala/dagr/core/cmdline/parsing/CommandLineParser.scala#L63)):_
Looking at ["effective scala" ](http://twitter.github.io/effectivescala/#Collections-Use), it is good practice to qualify a mutable collection.

---

_[core/src/main/scala/dagr/core/config/Configuration.scala, line 44 [r1]](https://reviewable.io:443/reviews/fulcrumgenomics/dagr/48#-K9n4rmG6Q7DTfFfAThG-r1-44:-K9sF59WDSTsobSt1iWd:210073) ([raw file](https://github.com/fulcrumgenomics/dagr/blob/c29b2cb93667b7b5157f57599422b6aa1383f129/core/src/main/scala/dagr/core/config/Configuration.scala#L44)):_
I will move out `Keys` since it is useful to folks outside of dagr, and make Configuration `private[core]`.

---

_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/fulcrumgenomics/dagr/48#-:-K9sGxthZlg4iumjJXxm:-835991766)_

tfenne:
Stupid Reviewable.  My responses to various points:
1. I will have to read through that "effective scala" link and see how much I agree with.  I would agree that for any collection type that exists in both mutable and immutable that's true, but ListBuffer only exists in mutable and is unambiguous.
2. Regarding `Configuration.Keys` I don't think that needs to be exposed outside of core, that's part of _why_ I nested it inside the Configuration object.  Can we just leave that as it was please?
3. ClpReflectiveBuilder extends it to deal with all our CLP annotations, ReflectiveBuilder is intended to be standalone.

nh13:
@tfenne could we try reviewable for any new PR.  My apologies for going their mid-review.  I will make the changes, and merge.

Issue #34, opened 2016-02-05
Unit test intermittently fails: "resubmit a task that failed and re-run to completion"


nh13:
```
[info] - should resubmit a task that failed and re-run to completion *** FAILED ***
[info]   FAILED_COMMAND was not equal to SUCCEEDED (TaskManagerTest.scala:228)
```

Note the line number may change given additional tests, so use the test name.

nh13:
fixed.

nh13:
It's still failing on some builds, so there must be another race condition. If this continues to be a problem, lets comment out this functionality and the test, since it is currently not being used.

nh13:
Turned off the tests and uncommented resubmit.

PR #49, opened 2016-02-05
fixing a bug where resubmitting a task may fail due the task not being

nh13:
properly removed from TaskRunner when it has completed.

[Review on Reviewable](https://reviewable.io/reviews/fulcrumgenomics/dagr/49)


nh13:
fixes #34

PR #50, opened 2016-02-06
Better support for piping tasks together

tfenne:
@nh13 Just putting this here in case you want to take a look and see how it's shaping up while I'm still cleaning up.  If you check out this branch the workflows all work with the new piping!


core/src/main/scala/dagr/core/tasksystem/PipedTask.scala
```diff
@@ -0,0 +1,123 @@
+package dagr.core.tasksystem
```
nh13: license

tasks/src/main/scala/dagr/tasks/misc/BwaMemStreamed.scala
```diff
@@ -53,48 +54,25 @@ class BwaMemStreamed(unmappedBam: PathToBam,
...
+
+    // Optionally tie in alt-processing if requested and the .alt file exists
     val alt = PathUtil.pathTo(ref + ".alt")
     if (processAltMappings && alt.toFile.exists()) {
-      tasks ++= new BwaK8AltProcessor(altFile = alt).getTasks
-      tasks ++= new FifoBuffer().requires(memory = Memory(fifoBufferMem)).getTasks
+      val k8       = new BwaK8AltProcessor(altFile = alt)
+      val k8Buffer = new FifoBuffer[Sam].requires(memory=Memory(fifoBufferMem))
+      pipe = pipe | k8 | k8Buffer
     }
```
nh13: It would be neat to be able to do something like this:

```
    val alt = PathUtil.pathTo(ref + ".alt")

    val samToFastq = SamToFastq(in=unmappedBam, out=Io.StdOut).requires(Cores(0.5), Memory(samToFastqMem))
    val fqBuffer   = new FifoBuffer[Fastq].requires(memory=Memory(fifoBufferMem))
    val bwaMem     = new BwaMem(fastq = Io.StdIn, ref = ref, minThreads = minThreads, maxThreads = maxThreads, memory = Memory(bwaMemMemory))
    val bwaBuffer  = new FifoBuffer[Sam].requires(memory=Memory(fifoBufferMem))

    // Optionally tie in alt-processing if requested and the .alt file exists
    var pipe = samToFastq | fqBuffer | bwaMem | bwaBuffer | if (processAltMappings && alt.toFile.exists()) {
      val k8       = new BwaK8AltProcessor(altFile = alt)
      val k8Buffer = new FifoBuffer[Sam].requires(memory=Memory(fifoBufferMem))
      k8 | k8Buffer
    } 
    else {
      PipeEnd
    }

    val mergeBam = new MergeBamAlignment(unmapped=unmappedBam, mapped=Io.StdIn, out=mappedBam, ref=ref).requires(memory=Memory(mergeBamAlignmentMem))
    root ==> (pipe | mergeBam).withName("BwaMemThroughMergeBam")
  }
```

I am drawing inspiration from the spray-routing DSL (ex. `PathEnd`)

tasks/src/main/scala/dagr/tasks/misc/BwaMemStreamed.scala
```diff
@@ -53,48 +54,25 @@ class BwaMemStreamed(unmappedBam: PathToBam,
                      mergeBamAlignmentMem: String = "2G",
                      fifoBufferMem: String = "512M",
                      processAltMappings : Boolean = true
-                    ) extends PipedTask with VariableResources {
-  this.name = "BwaMemStreamed"
-  override protected val tasks = ListBuffer[ProcessTask]()
+                    ) extends Pipeline {
```
nh13: Is there a reason not extend `Piping[Fastq,Sam]`?

tasks/src/main/scala/dagr/tasks/picard/FastqToUnmappedSam.scala
```diff
@@ -52,30 +53,23 @@ class FastqToUnmappedSam(fastq1: PathToFastq,
                          library: Option[String] = None,
                          readGroupName: Option[String] = None,
                          useSequentialFastqs: Boolean = false
-                          ) extends PipedTask with FixedResources {
+                          ) extends Pipeline {
```
nh13: Should all tasks really extend `Piping`?

tasks/src/main/scala/dagr/tasks/picard/MergeBamAlignment.scala
```diff
@@ -40,7 +42,8 @@ class MergeBamAlignment( unmapped: PathToBam,
                          attributesToRetain: List[String] = List[String]("X0", "ZS", "ZI", "ZM", "ZC", "ZN"),
                          orientation: String = "FR",
                          sortOrder: SortOrder = SortOrder.coordinate)
-  extends PicardTask {
+  extends PicardTask with Piping[SamOrBam,SamOrBam]
+{
```
nh13: cleanup

nh13:
@tfenne this is really cool.  This improves our DSL for writing piped tasks. I made a few comments about the use of `Piping`.  Let me know when I should do a full review.

tasks/src/main/scala/dagr/tasks/misc/BwaMemStreamed.scala
tfenne: I think this is a little complicated.  When we want to encapsulate a set of streamed tasks I think there are three options:
1. extend pipeline (as here).  You could, in theory, extend `Piping` and implement those methods to vector to the internally held piping object, but I think that violates some of our notions about depending across pipeline boundaries.
2. extend Task directly, extend 'Piping' and implement vectoring methods.  Basically the same as 1, but without violating `Pipeilne`'s contract
3. Don't have a class at all, but just an object with an apply method that returns the pipe chain.

The third option seems best to me now that I think about it.

tasks/src/main/scala/dagr/tasks/misc/BwaMemStreamed.scala
tfenne: Thanks!  I really like this idea, and was in fact struggling with how to write this section.  I'll give this a shot.  (We could also do this for ==> instead of using NoOpTask everywhere btw).

tasks/src/main/scala/dagr/tasks/picard/FastqToUnmappedSam.scala
tfenne: I think FastqToUnmappedSam should return something that is pipeable(!), but I don't think all tasks should extend `Piping`.  Metrics tasks for example could extend `PipeIn` but it seems like that would be over very limited utility since they would discard their input.

tfenne:
@nh13 I think this is now ready for a more in depth review.  I've settled on names for classes and traits (pending feedback).  I've implemented your suggestion, though given the type-checking for pipes it's a touch more verbose (`Pipes.empty[Sam]` vs. say `PipeEnd`), but is much better than the prior solution.

The only thing I'm not 100% happy with is that `Pipe` extends process task thus also `EmptyPipe` has to as well.  It's all hidden, and it's by no means the worst thing in the world, it just means it's allocating a bunch of stuff down in Task (a few `ListBuffer`s mainly) that there is no need for.  At some point in the future we could extract the `args` method from `ProcessTask` into an `Executable` trait which is mixed into both `ProcessTask` and `Pipe`, but I'm also concerned about creating a massive jumble of traits.

tfenne:
Oh, I should also say that I need to add a bunch of tests for this before final merge.

tfenne:
@nh13 I believe this is now ready for full review.  I'm finding Reviewable unpleasant - would you mind doing this one on GitHub?

tfenne:
@nh3 I'm finding the codecov report very useful!  I had forgotten in this case that I'd made some modifications to `Resource` and it nicely highlighted the new lines there that I had not tested.

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -0,0 +1,182 @@
...
+/**
+  * Trait for tasks that would like to be able to pipe data using stdin and stdout. Tasks
+  * are still responsible for managing their input and output - so if tasks can also read
+  * and write from files, they must have the input set to Io.Stdin and/or output to Io.Stdout
+  * before connecting them with pipes.
+  *
+  * See also the traits [[PipeIn]] and [[PipeOut]] which provide a simplified way to implement
+  * a piping task that only receives on or only emits to a pipe.
+  *
+
```
nh13: extra nl

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -0,0 +1,182 @@
...
+  * @tparam In a symbolic type representing the kind of data the task can accept
+  * @tparam Out a symbolic type representing the kind of fdata the task emits
```
nh13: typo `fdata`

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -0,0 +1,182 @@
...
+  */
+trait Pipe[In,Out] extends ProcessTask {
+  /**
+    * Generates or extends a pipe chain by connecting this task to the next task via a pipeline.
+    *
+    * @param next a Pipe enabled task that should receive this pipe's output as its input
+    * @tparam In2 the type of input expected by the next task. Must be the same or a supertype
+    *             of this pipe's output.
+    * @tparam Out2 the type of output of the next task, which becomes the output type of the
+    *              returned pipe chain
```
nh13: period

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -0,0 +1,182 @@
...
+  def |[In2 >: Out,Out2](next: Pipe[In2,Out2]): Pipe[In,Out2] = {
+    Pipes.chain(this, next)
+  }
+
+  /** Returns the first task in the pipe. */
+  private[tasksystem] def left: Pipe[In,_] = this
+  /** Returns the last task in the pipe. */
+  private[tasksystem] def right: Pipe[_,Out] = this
+  /** Returns the ordered set of Pipes/tasks in the chain from first to last inclusive. */
+  private[tasksystem] def ordered: Seq[Pipe[_,_]] = List[Pipe[_,_]](this)
```
nh13: why not return a `Seq` rather than a `List`?

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -0,0 +1,182 @@
...
+    override def args: Seq[Any] = throw new IllegalStateException("Should not reach here.")
+    override def pickResources(availableResources: ResourceSet): Option[ResourceSet] = throw new IllegalStateException("Should not reach here.")
+  }
+
+  /**
+    * An immutable class representing a chain of tasks that can be piped together. Currently only supports chains
+    * with up to a maximum of one non-FixedResources task, in order to keep the resource management code simple.
+    *
+    * The constructor does not check types as thoroughly as one might like, instead this is done in
+    * [[Pipes.chain()]], which is the only place the constructor is called from.
```
nh13: which ... from -> from which ...

nh13: I think it would be pretty easy to check in the constructor, given some modification to `Pipe`.  This works in my Repl:

```
import reflect.runtime.universe._
import scala.reflect.runtime.{universe => ru}
import scala.reflect._

val mirror: ru.Mirror = scala.reflect.runtime.currentMirror

class Pipe[Left: ClassTag, Right: ClassTag] {
  def left = classTag[Left]
  def right = classTag[Right]
}

def checkPipes(first: Pipe[_,_], second: Pipe[_,_]): Boolean = {
  first.right == second.left
}

val pipes = List[Pipe[_,_]](new Pipe[Int, String], new Pipe[String, Char], new Pipe[Char, Double], new Pipe[Int, Int])
for (i <- 1 until pipes.length) {
  println(s"Comparing ${i-1} and $i: " + checkPipes(pipes(i-1), pipes(i)).toString())
}
```

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -0,0 +1,182 @@
...
+      */
+    private def partition: (Seq[FixedResources], Option[ProcessTask]) = {
+      val (fixed, variable) = this.tasks.partition(task => task.isInstanceOf[FixedResources])
+      (fixed.map(_.asInstanceOf[FixedResources]), variable.headOption)
+    }
+
+
+    /**
+      * Picks resources for the piped set of tasks by attempting to allocate the necessary resources
+      * to all the tasks that implement FixedResources, and then offering the rest to the optional
```
nh13: in comments, if you reference code or classes, could you use "`code`" or `[[ClassName]]`?

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -0,0 +1,182 @@
...
+        val remaining = availableResources - fixedResources
+        variable.flatMap(_.pickResources(remaining)) match {
+          case Some(resources) => Some(fixedResources + resources)
+          case None => Some(fixedResources)
+        }
+      }
+      else {
+        None
+      }
+    }
```
nh13: why not just 

```
availableResources.subset(fixedResources).map { remaining =>
   variable.flatMap(_.pickResources(remaining)) match {
     case Some(resources) => Some(fixedResources + resources)
     case None => Some(fixedResources)
   }
}
```

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -0,0 +1,182 @@
...
+    override def applyResources(resources: ResourceSet): Unit = {
+      val (fixed, variable) = partition
+      var remainingResource = resources
+      fixed.foreach(task => { task.applyResources(task.resources); remainingResource -= task.resources })
+      variable.foreach(task => task.applyResources(remainingResource))
+    }
+  }
+
+  /** Returns an "empty" pipe that can be pushed into a pipe chain and does nothing. */
+  def empty[T]: Pipe[T,T] = new EmptyPipe[T]
```
nh13, line 149: thanks!

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -0,0 +1,182 @@
...
+      case _ =>
+        right match {
+          case p: EmptyPipe[Out2] => left.asInstanceOf[Pipe[In1,Out2]]
+          case _ =>
+            val xs = ListBuffer[Pipe[_,_]]()
+            xs.appendAll(left.ordered)
+            xs.appendAll(right.ordered)
+            new PipeChain(tasks=xs.toList, first=left.left, last=right.right)
+        }
+    }
```
nh13:
```
(left, right) match {
  case (p: EmptyPipe[Out1], _) => right.asInstanceOf[Pipe[In1,Out2]]
  case (_, p: EmptyPipe[Out2]) => left.asInstanceOf[Pipe[In1,Out2]]
  case (_, _)                  => new PipeChain(tasks=(left.ordered ++ right.ordered), first=left.left, last=right.right)
}
```

core/src/test/scala/dagr/core/tasksystem/PipeTest.scala
```diff
@@ -0,0 +1,132 @@
...
+    pipe.args.mkString(" ") shouldBe """cat foo.txt | sed s/ +/,/g | column -t"""
+  }
+
+  it should "compile with a subtype as input in a pipe" in {
+    val pipe1 = Cat("foo.txt") | BgZip() // exact type: Text
+    val pipe2 = Cat("foo.txt") | MakeCsv() | BgZip() // subtype: Csv
+    val pipe3 = Cat("foo.txt") | MakeCsv() | CsvToTsv() | BgZip() // subtype: Tsv
+  }
+
+  "EmptyPipe" should "explode if it's arg or pickResources methods are ever called" in {
```
nh13: `it's` -> `its`

tasks/src/main/scala/dagr/tasks/bwa/Bwa.scala
```diff
@@ -0,0 +1,62 @@
...
+    val alt = PathUtil.pathTo(ref + ".alt")
+    val doAlt = Files.exists(alt) && processAltMappings
+    val fifoMem: Memory = Memory(fifoBufferMem)
+
+    val samToFastq = SamToFastq(in=unmappedBam, out=Io.StdOut).requires(Cores(0.5), Memory(samToFastqMem))
+    val fqBuffer   = new FifoBuffer[Fastq].requires(memory=fifoMem)
+    val bwaMem     = new BwaMem(fastq=Io.StdIn, ref=ref, minThreads=minThreads, maxThreads=maxThreads, memory=Memory(bwaMemMemory))
+    val bwaBuffer  = new FifoBuffer[Sam].requires(memory=fifoMem)
+    val altPipe =  if (doAlt) new BwaK8AltProcessor(altFile=alt) | new FifoBuffer[Sam].requires(memory=fifoMem) else Pipes.empty[Sam]
+    val mergeBam = new MergeBamAlignment(unmapped=unmappedBam, mapped=Io.StdIn, out=mappedBam, ref=ref, sortOrder=sortOrder).requires(memory=Memory(mergeBamAlignmentMem))
```
nh13: align the `=`

tasks/src/main/scala/dagr/tasks/bwa/Bwa.scala
```diff
@@ -0,0 +1,62 @@
...
+    * @param ref the reference FASTA file, with BWA indexes pre-built
+    * @param sortOrder the desired sort order of the output BAM file
+    * @param minThreads the minimum number of threads to allocate to bwa-mem
+    * @param maxThreads the maximum number of threads to allocate to bwa-mem
+    * @param bwaMemMemory the amount of memory based on the referenceFasta
+    * @param samToFastqMem the amount of memory for SamToFastq
+    * @param mergeBamAlignmentMem the amount of memory for MergeBamAlignment
+    * @param fifoBufferMem the amount of memory for each FifoBuffer
+    */
+  def bwaMemStreamed(unmappedBam: PathToBam,
```
nh13: `bwaMemStreamed ` -> `BwaMemStreamed `

nh13, line 38: clever!

tasks/src/main/scala/dagr/tasks/bwa/Bwa.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    val samToFastq = SamToFastq(in=unmappedBam, out=Io.StdOut).requires(Cores(0.5), Memory(samToFastqMem))
+    val fqBuffer   = new FifoBuffer[Fastq].requires(memory=fifoMem)
+    val bwaMem     = new BwaMem(fastq=Io.StdIn, ref=ref, minThreads=minThreads, maxThreads=maxThreads, memory=Memory(bwaMemMemory))
+    val bwaBuffer  = new FifoBuffer[Sam].requires(memory=fifoMem)
+    val altPipe =  if (doAlt) new BwaK8AltProcessor(altFile=alt) | new FifoBuffer[Sam].requires(memory=fifoMem) else Pipes.empty[Sam]
+    val mergeBam = new MergeBamAlignment(unmapped=unmappedBam, mapped=Io.StdIn, out=mappedBam, ref=ref, sortOrder=sortOrder).requires(memory=Memory(mergeBamAlignmentMem))
+    (samToFastq | fqBuffer | bwaMem | bwaBuffer | altPipe | mergeBam).withName("BwaMemThroughMergeBam")
+  }
+}
```
nh13: this is glorious

nh13:
Overall, this greatly improves how to write piped tasks in dagr, thank-you.  Could you give an honest attempt at verifying the types of the pipes match in the constructor for `PipeChain`?  I think my example could be applied.  Other than that, some nitpick comments and then you should be ready to go.

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
tfenne: Because I always forget that you can do `Seq(…)`.  Fixed.

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
tfenne: I will try.  I've made a pass through Pipe.scala and fixed everything I could see.

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
tfenne: slowclap

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
tfenne: There are a few problems with this:
1. subset returns the subset, not the remaining amount after removing the subset
2. if there is no variable resource, `variable.flatMap` will simply return None, which is wrong
3. I think with optional resource sets and optional tasks, it's very hard to decode the intent (hence 2 above).

I think this is both shorter and clearer (with the introduction of a new method in ResourceSet that does what you think subset should do):

``` scala
variableTask match {
  case None    => availableResources.subset(fixedResources)
  case Some(v) => availableResources.minusOption(fixedResources) flatMap { remaining =>
    v.pickResources(remaining) map (_ + fixedResources)
  }
}
```

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
tfenne: So, I'm predisposed against doing this for a couple of reasons:
1. My comment here was not that things are un-typesafe, just that the typesafety is enforced elsewhere, so we need to be a touch more careful editing this code not to break type safety by making this class or constructor non-private
2. Requiring `ClassTag`s or `TypeTags` makes life more difficult for users.  It's infectious: any code that uses type parameters and your API, also has to require `TypeTag`s, so they'll be available when your API is called.  E.g. we'd have to rewrite `FifoBuffer` to also take a TypeTag since it's generic and doesn't care about it's data type.  Then in client code you couldn't write:

``` scala
def buffer[T] = new FifoBuffer().requires(memory="1G").withName("buffer" + counter++)
```

because you'd get some esoteric compiler warnings about TypeTags, and then you'd have to write:

``` scala
def buffer[T: TypeTag] = new FifoBuffer().requires(memory="1G").withName("buffer" + counter++)
```
1. The types are all checked by the compiler as it stands (try it!). I don't think runtime enforcement is necessary.  If you cast things and do bad things, that's you're problem.  E.g. try this in the REPL:

``` scala
val xs = ListBuffer[String]()
val ys = xs.asInstanceOf[ListBuffer[Any]]
ys.append(1)
xs foreach ((x: String) => println(x))
```

It compiles just find and gives a cast exception at runtime.

tfenne:
@nh13 I've fixed or responded to all your comments.  I'd like to squash and merge.  Any objections?

nh13:
one final nitpick and you should be ready to squash and merge.

PR #56, opened 2016-02-07
DO NOT MERGE: playing around with implementing scatter gather

nh13:


review-ninja:
<a href="https://app.review.ninja/fulcrumgenomics/dagr/pull/56" target="_blank">ReviewNinja

nh13:
@tfenne I am not sure why review-ninja is still commenting.

tfenne:
@nh13 I started reading, and I'll come back again later, but I'm struggling a little to understand what's going on.  It would be helpful, as you suggest, to have a real example.  Or, at least a cartoon of one that helps me understand the design goals of the machinery you have.

nh13:
@tfenne let me put an example up, and you can review then.

nh13:
@tfenne please see the `ScatterGatherPipelineTest` for examples.

This scatter gather framework and set of pipelines facilitate a lot of flexibility for scatter gather.  One thing that could make it even more flexible would be to have the output type of the scatter step be different than the output type of the gather step.  

Nonetheless, it is likely that a lot of real-world uses of this framework will want the same type for `Input` and `Intermediate`, for example, we split an interval list into multiple sub interval lists, and call or merge variants on the regions in each sub-list in the scatter.  I am hesitant to make this simplification in the base trait since I like the flexibility, or add it as a convenience trait/class since I want to keep the code short for this first review.  I did, however, add some apply methods to enable folks who don't want to implement the various traits but simply want to supply pure methods.  In this case, all the split, scatter, and gather tasks will run in the JVM.

There's some complexity that I couldn't get rid of in the code due to having to make sure that the outputs of each phase (split the input and scatter) are fully populated before we create the tasks involved with the next.  For example, we need to split the input before we can create the scatter tasks.  This requires wrapping some tasks in other tasks.

If you aren't grokking this code, lets chat in person.

core/src/main/scala/dagr/core/tasksystem/ScatterGatherFramework.scala
```diff
@@ -0,0 +1,116 @@
...
+  * [[ScatterTask]]) are responsible for step (2), and a gather task (see [[GatherTask]]) is responsible for step (3).
+  *
+  * The public methods are the minimum that need to be implemented to use the scatter gather framework.
+  *
+  * Override the protected methods to customize how scatter or gather tasks are generated, for example to perform
+  * recursive merging (ala merge-sort).  Otherwise, mix in traits that do this for you.
+  *
+  * Be sure to mix-in the appropriate trait in the task returned by each method.
+  *
+  * @tparam Input        the input type to the scatter gather.
```
tfenne: I really like `Domain` for this.  I think part of the reason I found this confusing the first time is that for scattering there are really two inputs:
1. The actual data (e.g. a BAM file, a VCF, some other large datasets)
2. The domain in which you are partitioning/scattering the data, e.g. genomic intervals

Since the pattern tends to be applied to indexed or otherwise sub-accessible data, the partitioning is applied to the domain and the scattering is performed by the scatter tasks.

Using a name like `Input` makes me thing we're talking about the bulky data (like BAM or VCF) and not the domain in which the partitioning is done.

core/src/main/scala/dagr/core/tasksystem/ScatterGatherFramework.scala
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/** All tasks that split the input into inputs for the scatter tasks should extend this trait.
+  *
+  * Provides a method that returns the inputs to the scatter tasks.  The method
+  * should only be called *after* this task has been run.  This is typically ensured by
+  * using a [[Callback]]. */
+trait SplitInputTask[Input,Intermediate] extends Task {
+  protected var _inputs: Option[Iterable[Intermediate]] = None
+  final def inputs: Iterable[Intermediate] = _inputs.get
```
tfenne: I'd be wary of using function names like `input`, `inputs`, `output` etc. in these traits, as that's going to make it really hard to mix them into task classes that have `val`s with those names.

I think the ideal usage of these traits will be something like:

``` scala
new IntervalListTools(....) with SplitInputTask {
  def scatteredInputs = ...
}
```

tfenne:
Thanks for updating this - I find it much clearer now.  I read through and feel like I totally understand `ScatterGatherFramework.scala` and tests.  I skimmed `ScatterGatherPipeline.scala` because I don't think it's as critical and I also found it a touch confusing still (just trying to grok your factorization of things).

An in person discussion would probably help hammer out the details fastest, but here are a few general thoughts:
1. There's a temptation here to make this into a general map/reduce framework vs. a scatter/gather framework.  The difference IMHO is that a map/reduce breaks apart an inputs, processes each piece independently and the reduces the output, where a scatter/gather framework partitions a domain of a dataset (intervals on a genome, byte ranges in a file), independently processes sub-domains or a single input and merges those outputs back together.  So while scatter/gather can be expressed within map/reduce, I think a simpler implementation/model is possible when restricting to just scatter/gather.
2. If you go a step further and limit it to working on `File`s or `Path`s you can also do all of the temporary file and directory management in the pipeline/framework.  E.g. if scattering HaplotypeCaller you probably want a directory per scatter to write files into per-task, and then to clean them up post-merge.  It would also allow you to _specify_ the outputs for scatter and gather tasks, instead of making them tell the framework.
3. I'm torn on the use of `Task` vs. method for the splitting of inputs.  I think what you've done is probably the best, since it gives the flexibility of using a Task to do the splitting and provides convenience mechanisms for doing it with just a function.  But I also feel like just using a function here would simplify things quite a bit, and let you get rid of some lazy-task stuff etc.  Since all execution is local in dagr, the only thing this takes away is the ease of running an external process to aid in the scattering - which you could still do, just it requires a touch more work.
4. Does this belong in core?  I think it should probably live in `tasks`, for a couple of reasons.  Firstly, I don't think it needs to be in core, and I don't _think_ it's likely that any of this would need tighter integration with any of the pieces of core.  Secondly, having it in core implies to me that it couldn't easily be implemented differently by someone else outside of core.  I asked myself the same question around the `Pipe` stuff, and wavered a bunch.  I ended up leaving that in core because I felt that if we change the execution model for ProcessTasks, we might want to move some of the logic for joining process tasks together into parts of core.

nh13:
Just so I remember, the next step is to create a framework to implement File-based scatter gather, that simplifies this framework, and does what is outlined above.  I don't think this above framework is truly map reduce, so perhaps it should remain named `scatter/gather`?  @tfenne what do you think

tfenne:
If it were me, I'd simplify the whole thing down to working with Paths and call it a day :wink: But yes, I think the next step is to see if it can be done in two layers - the very general map/reduce-ish framework you have now, and then a more concrete scatter/gather framework that specializes it and simplifies handling of all the temporary files and directories.

nh13:
@tfenne I would appreciate a brief look at how I used this framework to parallelize germline variant calling using GATK (see `ParallelHaplotypeCaller`).  I am looking for high-level feedback, after which, I can go back and spend the time to document and clean up.

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+  @arg(flag="o", doc="The output VCF to which to write variants.")                  val vcf: PathToVcf,
+  @arg(flag="d", doc="Path to dbSNP VCF (for GenotypeGvcfs).")                      val dbsnp: Option[PathToVcf] = None,
+  @arg(flag="f", doc="Filter variants using Picard FilterVcf.")                     val filterVcf: Boolean = false,
+  @arg(flag="s", doc="The size of the region to call for each parallel task.")      val regionSize: Int = 25000000,
+  @arg(flag="t", doc="Temporary directory in which to store intermediate results.") val tmpDirectory: Option[Path]
+)
+extends PathBasedScatterGatherPipeline(
+  input                       = ref,
+  output                      = vcf,
+  splitInputPathTaskGenerator = (tmpDir: Option[Path]) => (ref: Path)             => new SplitIntervalsForCallingRegions(ref=ref, intervals=intervals, output=tmpDir, regionSize=regionSize),
```
nh13: perhaps call these `split`, `map`, and `reduce`?

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+import dagr.tasks.DagrDef._
+import dagr.tasks.gatk.{GenotypeGvcfs, HaplotypeCaller, SplitIntervalsForCallingRegions}
+import dagr.tasks.misc.DeleteFiles
+import dagr.tasks.parallel.PathBasedScatterGatherPipeline
+import dagr.tasks.parallel.PathBasedScatterGatherPipeline.{GatherPathTask, ScatterPathTask}
+import dagr.tasks.picard.{FilterVcf, GatherVcfs}
+
+object ParallelHaplotypeCaller {
+
+  /** Runs GenotypeGvcfs, HaplotypeCaller, and optionally FilterVcf on a given interval list. */
```
tfenne: Perhaps list these in order?

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+        val del = new DeleteFiles(gvcf)
+        hc ==> gt ==> del
+        List(hc, gt, del)
+      }
+    }
+  }
+}
+
+// TODO: could also use IntervalListTools to break it up n-ways
+// TODO: could also take in an interval list rather than use the whole genome
```
tfenne: This is done, right?

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+
+object ParallelHaplotypeCaller {
+
+  /** Runs GenotypeGvcfs, HaplotypeCaller, and optionally FilterVcf on a given interval list. */
+  private class VariantCallTask(ref: PathToFasta,
+                                bam: PathToBam,
+                                filterVcf: Boolean = true,
+                                dbsnp: Option[PathToVcf] = None,
+                                tmpDirectory: Option[Path] = None,
+                                intervals: PathToIntervals) extends ScatterPathTask {
```
tfenne: Is it not possible for this to `extends Pipeline with ScatterPathTask`?

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+import dagr.tasks.gatk.{GenotypeGvcfs, HaplotypeCaller, SplitIntervalsForCallingRegions}
+import dagr.tasks.misc.DeleteFiles
+import dagr.tasks.parallel.PathBasedScatterGatherPipeline
+import dagr.tasks.parallel.PathBasedScatterGatherPipeline.{GatherPathTask, ScatterPathTask}
+import dagr.tasks.picard.{FilterVcf, GatherVcfs}
+
+object ParallelHaplotypeCaller {
+
+  /** Runs GenotypeGvcfs, HaplotypeCaller, and optionally FilterVcf on a given interval list. */
+  private class VariantCallTask(ref: PathToFasta,
```
tfenne: I would call this something clearer, like ScatteredSingleSampleHaplotypeCaller

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+
+    private def makeTempFile(prefix: String): PathToVcf = tmpDirectory match {
+      case Some(tmpDir) => Files.createTempFile(tmpDir, prefix, ".vcf.gz")
+      case None => Files.createTempFile(prefix, ".vcf.gz")
+    }
+
+    def getTasks: Traversable[_ <: Task] = {
+      val Seq(gvcf: PathToVcf, finalVcf: PathToVcf) = Seq("gvcf.", "finalVcf.").map (makeTempFile)
+      val hc    = new HaplotypeCaller(ref=ref, intervals=Some(intervals), bam=bam, vcf=gvcf)
+      val gtGen = (vcf: PathToVcf) => GenotypeGvcfs(ref=ref, intervals=Some(intervals), gvcf=gvcf, vcf=vcf, dbSnpVcf=dbsnp)
```
tfenne, line 58: I'd name this like task, since it's a task generator, e.g. `GenotypeThisVcf`

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+object ParallelHaplotypeCaller {
+
+  /** Runs GenotypeGvcfs, HaplotypeCaller, and optionally FilterVcf on a given interval list. */
+  private class VariantCallTask(ref: PathToFasta,
+                                bam: PathToBam,
+                                filterVcf: Boolean = true,
+                                dbsnp: Option[PathToVcf] = None,
+                                tmpDirectory: Option[Path] = None,
+                                intervals: PathToIntervals) extends ScatterPathTask {
+    var _output: Option[Path] = None
```
tfenne, line 50: Why not have this be a val and initialize it inline instead of in `getTasks`?  Or just have it specified by the caller?

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+  |Calls germline variants in a single sample and then optionally hard filters them.
+  |
+  |Runs GenotypeGvcfs, HaplotypeCaller, and optionally FilterVcf in parallel by splitting the interval list into
+  |multiple non-overlapping regions.  If no interval list is given, runs on the entire reference.
+  """,
+  group = classOf[Pipelines]
+)
+class ParallelHaplotypeCaller
+( @arg(flag="r", doc="Path to the reference FASTA.")                                val ref: PathToFasta,
+  @arg(flag="i", doc="The input BAM file (indexed) from which to call variants.")   val bam: PathToBam,
```
tfenne: `bam/i` maybe call it `input`?

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+  @arg(flag="l", doc="Path to interval list of regions to call.")                   val intervals: Option[PathToIntervals] = None,
+  @arg(flag="o", doc="The output VCF to which to write variants.")                  val vcf: PathToVcf,
```
tfenne: `vcf/o` maybe call it `output`?

tasks/src/main/scala/dagr/tasks/parallel/PathBasedScatterGatherPipeline.scala
```diff
@@ -0,0 +1,142 @@
...
+
+import dagr.core.tasksystem.{Linker, Task}
+import dagr.tasks.misc.DeleteFiles
+
+object PathBasedScatterGatherPipeline {
+
+  /** All tasks that split inputs should extend this trait and implement the method to specify the outputs.  The latter
+    * method will only be called once this task has completed successfully.
+    */
+  trait SplitInputPathTask extends Task {
```
tfenne: Could this be achieved with generics instead of named traits?  E.g. `SplitInput[Path]`?

tasks/src/main/scala/dagr/tasks/parallel/PathBasedScatterGatherPipeline.scala
```diff
@@ -0,0 +1,142 @@
...
+    * The latter method will only be called once this task has completed successfully.
+    */
+  trait ScatterPathTask extends Task {
+    def output: Path
+  }
+
+  /** All tasks that gather multiple outputs into a single output should implement this task. */
+  trait GatherPathTask extends Task
+
+  /** A wrapper for the split input task. */
```
tfenne: This much is obvious from the name - but for what purpose?

tasks/src/main/scala/dagr/tasks/parallel/PathBasedScatterGatherPipeline.scala
```diff
@@ -0,0 +1,142 @@
...
+                                         ) extends SplitInputTask[Path, Path] {
+    def getTasks: Traversable[_ <: Task] = {
+      val splitInputTask = splitInputPathTaskGenerator(input)
+      val linker = Linker(from=splitInputTask, to=this)((from, to) => to._subDomains = Some(from.outputs))
+      splitInputTask ==> linker
+      List(splitInputTask, linker)
+    }
+  }
+
+  /** A wrapper for the scatter task. */
```
tfenne, line 35: This much is obvious from the name - but for what purpose?

tasks/src/main/scala/dagr/tasks/parallel/PathBasedScatterGatherPipeline.scala
```diff
@@ -0,0 +1,142 @@
...
+
+/** A Scatter Gather Pipeline that operates on paths.  The input, intermediates, and output are all Paths.
+  *
+  * The constructor requires methods to generate the various task types.  The first path to the split input and
+  * scatter task are the temporary directory in which to operate, and the second argument is the input path.  The
+  * gather task takes a sequence of output paths to gather and the final output path.
+  */
+class PathBasedScatterGatherPipeline
+(
+  input: Path,
```
tfenne: I see why, but calling this `input` is confusing, in that the "input" to the HC pipeline is really the BAM, but from the S/G framework's perspective the BAM is irrelevant and the Intervals are what it's working on.  Maybe call this `domain`, or something along those lines?

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,111 @@
...
+  @arg(flag="d", doc="Path to dbSNP VCF (for GenotypeGvcfs).")                      val dbsnp: Option[PathToVcf] = None,
+  @arg(flag="f", doc="Filter variants using Picard FilterVcf.")                     val filterVcf: Boolean = false,
+  @arg(flag="s", doc="The size of the region to call for each parallel task.")      val regionSize: Int = 25000000,
+  @arg(flag="t", doc="Temporary directory in which to store intermediate results.") val tmpDirectory: Option[Path]
+)
+extends PathBasedScatterGatherPipeline(
+  input                       = ref,
+  output                      = vcf,
```
tfenne: Does the scatter/gather pipeline actually need to know about the output?

tasks/src/main/scala/dagr/tasks/gatk/SplitIntervalsForCallingRegions.scala
```diff
@@ -0,0 +1,111 @@
...
+                                      output: Option[DirPath],
+                                      regionSize: Int = 25000000,
+                                      prefix: String = "split_interval",
+                                      suffix: String = ".interval_list"
+                                     ) extends SimpleInJvmTask with SplitInputPathTask {
+  requires(1, "2g")
+  withName("SplitIntervalsForCallingRegions")
+
+  /** Stores the output interval lists, but will only be non-empty after this task has been run. */
+  val outputs = new ListBuffer[PathToIntervals]()
```
tfenne: Perhaps use an `Option[List[PathToIntervals]]` then so it's clear when it's ready vs. not, and there's not public mutable state?

tasks/src/main/scala/dagr/tasks/gatk/SplitIntervalsForCallingRegions.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  private def writeIntervals(intvs: Seq[Interval], idx: Int, header: SAMFileHeader): Unit = {
```
tfenne: `intvs`?  That's just arbitrary!

tasks/src/main/scala/dagr/tasks/gatk/SplitIntervalsForCallingRegions.scala
```diff
@@ -0,0 +1,111 @@
...
+import scala.collection.mutable.ListBuffer
+
+
+/** Creates a set of interval list files given a sequence dictionary.
+  *
+  * If no intervals are given, it will create one or more interval list files per reference sequence.
+  *
+  * Each input interval will be broken up into non-overlapping regions of at most `regionSize` size.
+  */
+class SplitIntervalsForCallingRegions(ref: PathToFasta,
```
tfenne, line 47: So, I'm not confident that the HC will do the right thing if fed regions that don't terminate at Ns or chromosome boundaries.  Specifically I worry that it will a) extend it's active region past the boundary and b) emit variants from it's entire active region.  This is why picard does the split-on-Ns.  Should probably at least test this with some large call set and verify that you don't get duplicate variant calls near any boundaries.

tasks/src/main/scala/dagr/tasks/gatk/SplitIntervalsForCallingRegions.scala
```diff
@@ -0,0 +1,111 @@
...
+                                      intervals: Option[PathToIntervals],
+                                      output: Option[DirPath],
+                                      regionSize: Int = 25000000,
```
tfenne: Two things: 1) this is a _max_ right?  You might emit interval lists that are smaller.  2) It's not really a region size, but a scatter size since you're saying this is the max bases in each scatter?  So maybe `maxBasesPerScatter`?

tasks/src/main/scala/dagr/tasks/gatk/SplitIntervalsForCallingRegions.scala
nh13, line 47: I suppose I could write a tool to filter out variants outside the given interval or not in the interval list.  This seems easy enough to do if I find there is a problem with duplicate variants.

nh13, line 47: Also, not that this is authoritative, but it looks like passing in `-L` has no edge-effects: http://gatkforums.broadinstitute.org/gatk/discussion/2385/scatter-gather-to-run-haplotypecaller-in-parallel

I will confirm on WGS sample.

tfenne:
I was thinking a little more about this overnight, and realized that neither this nor the simplified version in my gist actually solves one of my use cases.  Doh!  That use case is the multi-sample variant calling case where:
1) HC is run scattered on >= 1 samples
2) A sites-only VCF is then made for each scattered VCF
3) The sites-only files are gathered
4) VQSR is run on the gathered sites-only file
5) Each individual VCF from (1) is the run through ApplyVqsr
6) The VQSR'd VCFs are gathered

I haven't tried implementing this yet, but I'm envisaging an alternate approach that provides an API something like this:

``` scala
val scatter: Scatter[PathToIntervals] = Scatter(intervals => splitIntervals(intervals))
val callers: Scatter[HaplotypeCaller] = scatter.map(il => new HaplotypeCaller(..., regions=il))
val sites: Scatter[HaplotypeCaller] = callers.map(c => new MakeSitesOnly(in=c.vcf, out=...))
val gatherSites: Gather[GatherVcfs] = sites.gather(s => new GatherVcfs(in=s.map(_.out), out=...))
val vqsr = new Vqsr(in=gatherSites.out, out=vqsrOut)
val recal: Scatter[PathToVcf] = callers.map(c => new ApplyVqsr(in=c.vcf, out=...))
val finalGather: Gather[GatherVcfs] = recal.gather(rs => new GatherVcfs(in=rs.map(_.out), out=finalVcf))

root ==> scatter
gatherSites ==> vqsr ==> recal // add dependencies the scatter/gather can't know about
```

This feels _very_ scala-ish, and I think is more general while not requiring that any of the scatter or gather tasks implement any specific traits.  A few thoughts:
1. `map()` and `gather()` would both set dependencies internally so you don't have to do it
2. `Scatter` and `Gather` would implement `Dependable` and vector dependencies to the created tasks
3. `Scatter` could easily have multiple methods for things like `gatherAll`, `gatherHierarchichally(n: Int)` etc.

Thoughts?

tasks/src/main/scala/dagr/tasks/parallel/PathBasedScatterGatherPipeline.scala
nh13: The idea of these traits was to have a more intuitive name for the underlying paths than `subDomains` or the like.

tasks/src/main/scala/dagr/tasks/gatk/SplitIntervalsForCallingRegions.scala
nh13: removed, only now need in the `run` method.

nh13:
Consider the case where we don't know the list of intervals (how many, what they're named) until the `Scatter(intervals => splitIntervals(intervals))` step has run.  Then `callers` can't be made until `scatter` has been run by `Dagr`.  Similarly, we may not always know ahead of time the outputs of the `map` or `reduce` step in advance. This is the reason for the `ScatterGatherFramework` I created, to defer actually building the tasks until the inputs are known.

Also, think of a case where we take an interval list file (`PathToIntervalList`), split it into multiple interval list (`Seq[Interval]`), and map that to a VCF (`PathToVcf`), and reduce to a final VCF (`PathToVcf`).  This what FreeBayes will want.  I think your API would support that.

nh13:
Closing this in favor of https://github.com/fulcrumgenomics/dagr/pull/171

PR #61, opened 2016-02-09
making a shell of a README

nh13:


tfenne:
How does this relate to the larger documentation project?  Is the idea to edit this here then copy it to the doc repo?

nh13:
I figure we need something now, and the other documentation project can continue forward.

nh13:
We should have a link to the `Linker` task (when ready) in the section " List of features" in the sub-section "Mechanisms for passing state".  See https://github.com/fulcrumgenomics/dagr/pull/97

nh13:
for the example Pipeline, do we want to include the imports?

nh13:
@tfenne just a few more questions, and I'd lock to block on the `Linker` PR.

tfenne:
@nh13 I don't think we should include the imports etc. for the example pipeline - I think that's just boilerplate that will make it look more intimidating.  If you want we could provide a link into the actual file in the repo also?

nh13:
@tfenne all set, rebase and merge.

tfenne:
How did this become mine to rebase and merge?  This is your PR :stuck_out_tongue:

PR #64, opened 2016-02-09
Edits in place of review of nh_with_style

tfenne:


core/src/main/scala/dagr/core/cmdline/parsing/ClpReflectiveBuilder.scala
```diff
@@ -211,32 +206,13 @@ private[parsing] class ClpArgument(declaringClass: Class[_],
...
+    def size: Int = (seq orElse set) map (_.size) getOrElse collection.get.size()
 
-    def isEmpty: Boolean = {
-      (seq.isDefined, set.isDefined) match {
-        case (true, _) => seq.get.isEmpty
-        case (_, true) => set.get.isEmpty
-        case _ => collection.get.isEmpty
-      }
-    }
+    def isEmpty: Boolean = size == 0
```
nh13, line 211: I will change this to the style of `size` since some underlying implementations may not need to get an accurate count to determine emptiness.

core/src/main/scala/dagr/core/cmdline/parsing/ClpReflectiveBuilder.scala
```diff
@@ -177,15 +171,15 @@ private[parsing] class ClpArgument(declaringClass: Class[_],
...
-        value = ParsingUtil.constructFromString(this.argumentType, this.unitType, values:_*)
-        this.isSetByUser = true
-    }
+    if (isSetByUser) throw new IllegalStateException(s"Argument '$name' has already been set")
+    if (!isCollection && values.size > 1) throw new UserException(s"Argument '${this.names}' cannot be specified more than once.")
+
+    if (isFlag && values.isEmpty)
+      this.value = true
+    else
+      this.value = ParsingUtil.constructFromString(this.argumentType, this.unitType, values: _*)
```
nh13, line 180: I strongly dislike this style.  Either on the same line or with curly braces.

core/src/main/scala/dagr/core/cmdline/parsing/ClpReflectiveBuilder.scala
tfenne, line 211: Ah, good point.

core/src/main/scala/dagr/core/cmdline/parsing/ClpReflectiveBuilder.scala
tfenne, line 180: This makes me sad, and I hope you will grow to like it ;)  But until then I will try and avoid this.

nh13, line 180: I am usually open minded, but there are a few things about which I am not, and this is one of them.

PR #68, opened 2016-02-10
A first couple of PreDef-like methods with tests.

tfenne:


tfenne:
Thinking out loud - perhaps these should be in package `dagr` and not `dagr.core` (but still within the core project).

core/src/main/scala/dagr/core/DagrDef.scala
```diff
@@ -0,0 +1,59 @@
...
+  /** An exception that implies that code is unreachable. */
+  private class UnreachableException(message: String) extends IllegalStateException(message)
+
+  /**
+    * A terse way to throw an `UnreachableException` that can be used where any type is expected,
+    * e.g. `Option(thing) getOrElse unreachable("my thing is never null")`
+    *
+    * @param message an optional message
+    */
+  def unreachable(message: => String = ""): Nothing = throw new UnreachableException(message)
```
nh13: I confirmed this works on `getOrElse`, so this makes me really happy.

core/src/main/scala/dagr/core/DagrDef.scala
```diff
@@ -0,0 +1,59 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package dagr.core
+
+/**
+  * Object that is designed to be imported with `import DagrDef._` in any/all dagr classes
+  * much like the way that scala.PreDef is imported in all files automatically.
+  */
```
nh13: Perhaps a warning about not shoving things in here arbitrarily.

nh13:
:+1:

PR #71, opened 2016-02-11
Implementation of argparse style argument files.

tfenne:
@nh13 I decided one commented out feature was enough, so did something about the argument file parsing.  This mimics how python's argparse does it, which I really like.  Makes it super-simple to have arguments full of spaces and special characters.

I'm open to suggestions (or PRs!) for improvement.  I'm not totally happy with the implementation of `takeNextToken` in `ArgTokenizer` but with the mix of `Try`s and `Option`s, I couldn't make it work with just maps and flatmaps.

### DIFF
--- a/sopt/src/main/scala/dagr/sopt/ArgTokenizer.scala
+++ b/sopt/src/main/scala/dagr/sopt/ArgTokenizer.scala
@@ -24,8 +24,12 @@
 
 package dagr.sopt
 
+import java.io.IOException
+import java.nio.file.{Files, Paths}
 import java.util.NoSuchElementException
 
+import scala.collection.JavaConversions._
+import scala.collection.mutable
 import scala.util.{Failure, Success, Try}
 
 object ArgTokenizer {
@@ -41,13 +45,15 @@ object ArgTokenizer {
 
 /** A class to tokenize a sequence of strings prior to option parsing.  A series of tokens will
   * be available.  If any errors are encountered, a failure will be present instead of the token.  */
-class ArgTokenizer(args: TraversableOnce[String]) extends Iterator[Try[ArgTokenizer.Token]] {
+class ArgTokenizer(args: Seq[String],
+                   val argFilePrefix: Option[String] = None) extends Iterator[Try[ArgTokenizer.Token]] {
   import ArgTokenizer._
 
   private var nextToken: Option[Try[Token]] = None
-  private val iterator = args.toIterator
+  private val stack = mutable.Stack[String](args:_*)
 
-  def this(args: String*) = this(args.toTraversable)
+  /** Alternate constructor that supports var-arg syntax for testing. */
+  private[sopt] def this(args: String*) = this(args.toSeq, None)
 
   /** True if there are more tokens, false otherwise */
   override def hasNext(): Boolean = {
@@ -60,8 +66,7 @@ class ArgTokenizer(args: TraversableOnce[String]) extends Iterator[Try[ArgTokeni
     // bad user
     if (!hasNext()) throw new NoSuchElementException("Called 'next' when 'hasNext' is false")
     val tryVal = nextToken match {
-      case None =>
-        throw new NoSuchElementException("Called 'next' when 'hasNext' is false")
+      case None        => throw new NoSuchElementException("Called 'next' when 'hasNext' is false")
       case Some(value) => value
     }
     nextToken = None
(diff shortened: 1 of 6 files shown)


tfenne:
Note to self: codecov reminded me I need to add a test for a missing file and make sure the error is reported correctly.

sopt/src/main/scala/dagr/sopt/ArgTokenizer.scala
```diff
@@ -41,13 +44,15 @@ object ArgTokenizer {
 
 /** A class to tokenize a sequence of strings prior to option parsing.  A series of tokens will
   * be available.  If any errors are encountered, a failure will be present instead of the token.  */
-class ArgTokenizer(args: TraversableOnce[String]) extends Iterator[Try[ArgTokenizer.Token]] {
+class ArgTokenizer(val argFilePrefix: Option[String] = None,
+                   args: Seq[String]) extends Iterator[Try[ArgTokenizer.Token]] {
```
nh13: I prefer putting the optional parameter at the end when there are no varargs.

sopt/src/main/scala/dagr/sopt/ArgTokenizer.scala
```diff
@@ -41,13 +44,15 @@ object ArgTokenizer {
...
   import ArgTokenizer._
 
   private var nextToken: Option[Try[Token]] = None
-  private val iterator = args.toIterator
+  private val stack = scala.collection.mutable.Stack[String](args:_*)
```
nh13: I prefer putting `import scala.collection.mutable` in the list of imports and then putting `mutable.Stack...`.  It makes it very clear you are choosing `mutable` and `Stack` without having to read through `scala.collection`.

sopt/src/main/scala/dagr/sopt/ArgTokenizer.scala
```diff
@@ -69,19 +74,34 @@ class ArgTokenizer(args: TraversableOnce[String]) extends Iterator[Try[ArgTokeni
...
+    case (arg,  _) if arg.startsWith("-") => Some(convertSingleDash(arg.substring(1)))
+    case (arg, Some(pre)) if arg.startsWith(pre) =>
+      try {
+        this.stack.pushAll(loadArgumentFile(arg.drop(pre.length)).reverseIterator)
+        if (this.stack.nonEmpty) takeNextToken else None
       }
-    }
+      catch {
+        case ex: IOException => Some(Failure(ex))
+      }
```
nh13: Consider using a `Try` here, although you may not want to catch all exceptions.

sopt/src/test/scala/dagr/sopt/ArgTokenizerTest.scala
```diff
@@ -165,4 +178,56 @@ class ArgTokenizerTest extends UnitSpec {
...
+    val tokenizer = new ArgTokenizer(Some("@"), Seq("--zero", "@" + path3, "--six"))
+    tokenizer.next() shouldBe Success(ArgOption(name="zero"))
+    tokenizer.next() shouldBe Success(ArgOption(name="one"))
+    tokenizer.next() shouldBe Success(ArgOption(name="two"))
+    tokenizer.next() shouldBe Success(ArgOption(name="three"))
+    tokenizer.next() shouldBe Success(ArgOption(name="four"))
+    tokenizer.next() shouldBe Success(ArgOption(name="five"))
+    tokenizer.next() shouldBe Success(ArgOption(name="six"))
+  }
+
```
nh13: no

nh13:
This is going to be very useful.  I wonder if you could update the `README.md` to include a description of an arguments file.  Other than that, looks good.

sopt/src/main/scala/dagr/sopt/ArgTokenizer.scala
tfenne: Will fix.  This is a hangover from trying to make the Seq[String] and String\* constructors match each other.

sopt/src/main/scala/dagr/sopt/ArgTokenizer.scala
tfenne: This is the section I pointed out in the mean thread.  I tried all kinds of things here to use a `Try`, and got tied in knots.  What I wanted to do was have `loadArgumentFile` return a Try[List[String]] instead of throwing exception, but it's all complicated by the fact that
1. This method returns an `Option[Try[Token]]`
2. On line 94 I need to return a straight `None`  in the case where the arguments file is the last argument in the stack, and the arguments file turns out to be empty

It's beyond me (currently) to do this with `Try`, but I'd be glad of some help if you think it's doable.

tfenne:
@nh13 I've addressed your issues, save for the `Try` vs. `try`, and added the extra test for a missing arguments file.  I'd love some help with `Try` if you're up for it, otherwise I think this is ready to merge.

sopt/src/main/scala/dagr/sopt/ArgTokenizer.scala
nh13: I see two options:

```
Try {
  this.stack.pushAll(loadArgumentFile(arg.drop(pre.length)).reverseIterator)
}.map(t => if (this.stack.nonEmpty) takeNextToken else None).getOrElse(None)
```

or 

```
Try {
  this.stack.pushAll(loadArgumentFile(arg.drop(pre.length)).reverseIterator)
} match {
  case Success(success) => if (this.stack.nonEmpty) takeNextToken else None
  case Failure(failure) => None
}
```

tfenne: Thanks.  I think I was trying to be _too_ clever, and didn't think of just doing a secondary match.  I ended up with a modified version of your second option so that I can do this:

``` scala
// Now returns Try
private def loadArgumentFile(filename: String): Try[Seq[String]] = {
  Try { Files.readAllLines(Paths.get(filename)).map(s => s.trim).filter(s => s.nonEmpty).toSeq }
}

// That bit of the match
case (arg, Some(pre)) if arg.startsWith(pre) =>
  loadArgumentFile(arg.drop(pre.length)) match {
    case Failure(failure) => Some(Failure(failure))
    case Success(newArgs) =>
      this.stack.pushAll(newArgs.reverseIterator)
      if (this.stack.nonEmpty) takeNextToken else None
  }
```

Issue #67, opened 2016-02-10
Refactor TaskManagerState


nh13:
Current comments:
- Perhaps we should just have private[core] var taskId : Option[Int] = None on Task to simplify the book-keeping?
- The separation between `TaskExecutionInfo` and `GraphNode` is also inconvenient.  It adds a lot of methods and complexity.
- Could the `taskForId` methods be accessed like arrays, similar to how we can do this `someMap(key)` instead of `someMap(key).get`?

nh13:
merged

PR #72, opened 2016-02-11
Major refactor of the execsystem

nh13:
@tfenne targeted the `execsystem` and made some much needed changes.  There's more to do, but I thought I would stop here prior to any further major changes.


nh13:
See #67

core/src/main/scala/dagr/core/execsystem/GraphNode.scala
```diff
@@ -37,11 +39,10 @@ import scala.collection.mutable.ListBuffer
   *                         is used to track submission, start, and end dates for the parent node.  Typically parent nodes
   *                         are [[dagr.core.tasksystem.Pipeline]]s.
   */
-class GraphNode(val taskId: BigInt,
-                var task: Task,
+class GraphNode(var task: Task,
                 predecessorNodes: Traversable[GraphNode] = Nil,
                 var state: GraphNodeState.Value = GraphNodeState.PREDECESSORS_AND_UNEXPANDED,
-                val parent: Option[GraphNode] = None) {
+                val parent: Option[GraphNode] = None) extends BaseGraphNode {
```
tfenne: This is my first time reading some of the graph code, so I'm going to ask some basic questions.  I see this constructor takes both `predecessorNodes` and an optional `parent`.  Is it the case that predecessors are nodes which must be succeeded before this node executes, and parent is the node that generated this node during expansion?  If that's right, I might suggest `enclosingNode` instead of parent to make the distinction clearer.

core/src/main/scala/dagr/core/execsystem/GraphNode.scala
```diff
@@ -37,11 +39,10 @@ import scala.collection.mutable.ListBuffer
...
 
   private val _predecessors = new ListBuffer[GraphNode]()
   private val _predecessorsStatic = new ListBuffer[GraphNode]()
```
tfenne: This caught my eye even though you didn't change it.  It's odd that this is called `static` but is a mutable type, and is in fact mutated in the add methods.  I looked around and it seems like the only real usage outside of this class is in one test.  I wonder if it's even worth keeping around in that case?

core/src/main/scala/dagr/core/execsystem/TaskExecutionInfo.scala
```diff
@@ -42,7 +42,6 @@ import dagr.core.tasksystem.Task
  * @param attemptIndex the one-based count of attempts to run this task.
  */
 class TaskExecutionInfo(var task: Task,
-                        val id: BigInt,
                         var status: TaskStatus.Value,
```
tfenne, line 45: Just wondering out loud ... whether `status`, `script` and `logFile` might all also become `private[core] var`s in Task in some future refactoring?

core/src/main/scala/dagr/core/execsystem/TaskExecutionRunner.scala
```diff
@@ -47,7 +47,7 @@ private object TaskRunner {
     * @param script the location of the task's script
     * @param logFile the location of the task's log file
     */
-  class SingleTaskRunner(task: ProcessTask, script: Path, logFile: Path) extends TaskRunnable(task = task) {
+  class SingleTaskExecutionRunner(task: ProcessTask, script: Path, logFile: Path) extends TaskRunnable(task = task) {
```
tfenne: `ProcessTaskExecutionRunner`?

tfenne: It won't let me comment on the appropriate line since you didn't change it, but I think it would be clearer if on interruption of the thread you set the return value to some less commonly used value than 1 (e.g. 255) and also had a constant for that so it's clear elsewhere.

core/src/main/scala/dagr/core/execsystem/TaskExecutionRunner.scala
```diff
@@ -89,7 +89,7 @@ private object TaskRunner {
   }
 
   /** Simple class that sets the exit code to zero */
-  class NoOpTaskRunner(task: UnitTask) extends TaskRunnable(task = task) {
+  class NoOpTaskExecutionRunner(task: UnitTask) extends TaskRunnable(task = task) {
```
tfenne: Maybe rename this to `SimulationTaskRunner` or comment that it's used for simulation.  I had to go hunting to figure out why you'd want this.

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -635,20 +511,18 @@ class TaskManager(taskManagerResources: TaskManagerResources = TaskManagerDefaul
     logger.debug("runSchedulerOnce: finishing one round of execution")
     logger.debug("runSchedulerOnce: found " + runningTasks.size + " running tasks and " + tasksToSchedule.size + " tasks to schedule")
 
-    (readyTasks, tasksToSchedule.keys.toList, runningTasks.keys.toList, completedTasks.keys.map(taskId => taskValueFor(taskId)).toList)
+    (
+      readyTasks,
+      tasksToSchedule.keys.toList,
```
tfenne: Just wondering if the API for this method might be loosened to returning `Iterable` or `Traversable` for some of these, since it seems like the ordering is not important and then you could do this without generating temporary lists every iteration.

core/src/main/scala/dagr/core/execsystem/TaskManagerLike.scala
```diff
@@ -0,0 +1,177 @@
...
+import dagr.core.execsystem.TaskManagerLike.BaseInfoAndNode
+import dagr.core.tasksystem.Task
+import dagr.core.util.BiMap
+
+object TaskManagerLike {
+  abstract class BaseInfoAndNode(info: TaskExecutionInfo, node: BaseGraphNode)
+}
+
+/** A generic template for task managers */
+trait TaskManagerLike {
```
tfenne: `private[core]` or `private[execsystem]`?

core/src/main/scala/dagr/core/execsystem/TaskManagerLike.scala
```diff
@@ -0,0 +1,177 @@
...
+    * @param task the task.
+    * @return the task identifier, None if the task is not being managed.
+    */
+  def taskIdFor(task: Task): Option[BigInt] = task.id
+
+  /** Get the task identifiers for all tracked tasks.
+    *
+    * @return the task ids
+    */
+  def taskIdsFor(): Iterable[BigInt]
```
tfenne: Maybe just `taskIds` or `allTaskIds`

core/src/main/scala/dagr/core/execsystem/TaskManagerLike.scala
```diff
@@ -0,0 +1,177 @@
...
+    */
+  def apply(id: BigInt): Task
+
+  /** Gets the task execution info for the given task.  If the task is not
+    *  being tracked, throws a [[NoSuchElementException]].
+    *
+    * @param task the task to lookup.
+    * @return the associated task execution info.
+    */
+  def apply(task: Task): BaseInfoAndNode
```
tfenne: I don't love having these two different map-like `apply()` methods in the same class.  When reading through TaskManager and finding things like `this(this(id)).info` I had to stop and puzzle out what it was doing.  Could we keep the `apply(id): Task` method, but rename this to `infoFor(task): BaseInfoAndNode`?

core/src/main/scala/dagr/core/execsystem/TaskManagerLike.scala
```diff
@@ -0,0 +1,177 @@
...
+    * @return the task identifier.
+    */
+  def addTask(task: Task): BigInt
+
+  /** Adds tasks to be managed
+    *
+    * @param tasks the given tasks.
+    * @return the task identifiers.
+    */
+  def addTasks(tasks: Task*): List[BigInt]
```
tfenne: Could provide a default impl here that is just `(tasks map addTask).toList`

core/src/main/scala/dagr/core/execsystem/TaskManagerLike.scala
```diff
@@ -0,0 +1,177 @@
...
+  /** Run all tasks managed to either completion, failure, or inability to schedule.
+    *
+    * This will terminate tasks that were still running before returning in the case of failure or
+    * inability to schedule.
+    *
+    * @param sleepMilliseconds the time to wait in milliseconds to wait between trying to schedule tasks.
+    * @param timeout           the length of time in milliseconds to wait for running tasks to complete
+    * @return a bi-directional map from the set of tasks to their execution information.
+    */
+  def runToCompletion(sleepMilliseconds: Int = 1000, timeout: Int = 1000): BiMap[Task, TaskExecutionInfo]
```
tfenne: I feel like `sleepMilliseconds` is an implementation detail, not a core piece of the API.  E.g. imagine you implemented `TaskManager` using futures or actors, so you were immediately notified any time a task completed and scheduling could be tried again.  I think it would make more sense to move `sleepMilliseconds` to be a constructor param of the existing TaskManager implementation and then to remove it here and from `stepExecution`.

tfenne: Could `timeout` be either a `java.util.concurrent.TimeUnit` or a `scala.concurrent.duration.Duration`?  It might also be nice to make it an `Option` type so you can specify no timeout if you desire?

core/src/main/scala/dagr/core/execsystem/TaskManagerLike.scala
```diff
@@ -0,0 +1,177 @@
...
+    * 4. Schedule tasks and start running them.
+    *
+    * @param timeout the length of time in milliseconds to wait for running tasks to complete
+    * @return a tuple of:
+    *         (1) tasks that can be scheduled.
+    *         (2) tasks that were scheduled.
+    *         (3) tasks that are running prior to scheduling.
+    *         (4) the tasks that have completed prior to scheduling.
+    */
+  def stepExecution(timeout: Int = 1000): (List[Task], List[Task], List[Task], List[Task])
```
tfenne: Just re-iterating that if these could be `Iterable` or `Traversable` that might be friendlier.

core/src/main/scala/dagr/core/execsystem/TaskStatusReporter.scala
```diff
@@ -0,0 +1,122 @@
...
+import java.sql.Timestamp
+
+import dagr.core.tasksystem.Task
+import dagr.core.util.BiMap
+import dagr.core.util.StringUtil._
+
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+/** Provides implements a method to provide an execution report for a task tracker */
```
tfenne: Extra word

core/src/main/scala/dagr/core/execsystem/TaskStatusReporter.scala
```diff
@@ -0,0 +1,122 @@
...
+trait TaskStatusReporter extends TaskTracker {
+
```
tfenne: Extra newline!  It's a bingo!

core/src/main/scala/dagr/core/execsystem/TaskStatusReporter.scala
```diff
@@ -0,0 +1,122 @@
...
+  /** Gets the time stamp as a string (without Nanoseconds), or NA if it is None */
+  private def timestampStringOrNA(timestamp: Option[Timestamp]): String = {
```
tfenne: Putting this here because I wasn't sure where else to put it.  Java 8 introduced a whole new API for dates/times/durations that is, I believe, modeled on Joda Time.  The overview doc is [here](https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html#package.description).  I'm not sure if you want to take in on as part of this PR, but I think it would be good to switch over to that vs. Timestamp sooner rather than later.

core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
```diff
@@ -0,0 +1,344 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 Fulcrum Genomics LLC
```
tfenne: 2016 or 2015-2016

core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
```diff
@@ -0,0 +1,344 @@
...
+  case class InfoAndNode(info: TaskExecutionInfo, node: GraphNode) extends BaseInfoAndNode(info=info, node=node)
+}
+
+/** Tracks tasks during their execution */
+trait TaskTracker extends TaskManagerLike with LazyLogging {
+
+  /** Stores information about the execution of the current task and its place in the execution graph.  Mainly used
+    * for convenience during lookups.
+    */
+  private final class TrackingInfo(task: Task, parent: Option[GraphNode] = None) {
```
tfenne: Is this parent in the sense of the "enclosing" graph node, or is this just the corresponding graph node for the task?

core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
```diff
@@ -0,0 +1,344 @@
...
+      val node = predecessorsOf(task = task) match {
+        case None => new GraphNode(task=task, predecessorNodes=Nil, state=GraphNodeState.ORPHAN, parent=parent)
+        case Some(predecessors) => new GraphNode(task=task, predecessorNodes=predecessors, parent=parent)
+      }
+      (info, node)
+    }
+  }
+
+  /** The next id for a task */
+  private var nextId: BigInt = 0
```
tfenne: Int or Long wasn't enough? :astonished:

tfenne: Also, since it's used in a lot of places it might make sense to have a `type TaskId = BigInt`, which would make it easy to change in the future, and also clearer what it is everywhere.

core/src/main/scala/dagr/core/tasksystem/Task.scala
```diff
@@ -154,6 +154,9 @@ trait Task extends Dependable {
   // A string set by the enclosing workflow to distinguish the task
   private[tasksystem] var contextName : Option[String] = None
 
+  /** The unique id given to this task by the execution system, or None if not being tracked. */
+  private[core] var id : Option[BigInt] = None
```
tfenne: Just wondering if this should be `_id` or `taskId`, so as not to prevent concrete task classes from defining their own property called `id` in the future?

tfenne: Nevermind, apparently that's not an issue.  This works fine in the REPL:

``` scala
class Foo { private var x: Int = 1 }
class Bar(val x: String) extends Foo 
```

tfenne: Doh, I take it back.  Because it's `private[core]` and not just `private` this is in fact an issue.

tfenne:
This looks much nicer @nh13.  I have to admit I'm still a bit confused about _why_ the functionality for TaskManager is broken into multiple traits - it doesn't seem like they are intended to be mixed into other classes or composed in different ways, and they're not all that independent:

``` scala
trait TaskManagerLike
trait TaskTracker extends TaskManagerLike
trait TaskStatusReporter extends TaskTracker
class TaskManager(...) extends TaskManagerLike with TaskTracker with TaskStatusReporter
```

I think it's mostly so as not to have a single ~1000 line class, and to try and separate some of the responsibilities?  If so it would be nice to work, longer term, towards something like:

``` scala
// Just the public API methods that should be exposed outside of execsystem
// Entirely abstract except for methods that map collections onto other methods in the trait
trait TaskManagerLike

// Purely private state and protected methods for manipulating it
// Doesn't extend TaskManagerLike or implement any methods from it
// methods like `addTask` and `replaceTask` which have a lot of implementation logic
// move down into TaskManager
trait TaskTracker

// Adds reporting functionality to any class that is a TaskTracker, but doesn't extend
trait TaskStatusReporter { 
  this: TaskTracker =>
}

// Is the first time the three traits ever really come together
class TaskManager(...) extends TaskManagerLike with TaskTracker with TaskStatusReporter
```

core/src/main/scala/dagr/core/execsystem/GraphNode.scala
nh13: You're interpretation is right.

core/src/main/scala/dagr/core/execsystem/GraphNode.scala
nh13: I'll add it back in when I need it, which I anticipate being when I need to have a copy of the original execution graph; see scaladoc for `originalPredecessors`.  The other option is to have `_predecessors` be a `Map[GraphNode,Boolean` storing `true/false` for a predecessor if it has been satisfied.  I suppose I can make the latter change at a later date.  Thoughts?

tfenne: I see.  My tendency is nearly always to implement only what's needed now, and let future me figure out the best way to support future features, since it means that the system is as simple as it can be at any time.

core/src/main/scala/dagr/core/execsystem/TaskExecutionInfo.scala
nh13, line 45: I wanted task at being good at the execution part, not the tracking part, hence another class to store all of this info.  You could ask the same question about the other info too.  If we go down the route of putting more of this info into task, I would suggest having task store a reference to a `TaskExecutionInfo`.

tfenne, line 45: I see.  I think putting a reference to `TaskExecutionInfo` into `Task` makes sense since it would simplify more of the tracking/book-keeping.  The other option would be to use `TaskExecutionInfo` _everywhere_ in TaskManager and friends, and have it be the _only_ way to get at a _Task_, so that it's basically a super-wrapper around the task inside the execution/management world.

I find what I really want is post-creation ad-hoc mixins:
`val superTask = myTask with TaskExecutionInfo`
so that you could augment an _instance_ you already have with another trait.

core/src/main/scala/dagr/core/execsystem/TaskManagerLike.scala
nh13: I'll also make it a `Seq`, since I want to return the ids in the same order, and is a bit more general than `List`.

core/src/main/scala/dagr/core/execsystem/TaskManagerLike.scala
nh13: TaskExecutionRunner is up next for refactor (I want to make a trait, and have a single implementation), so I'll add a TODO.

core/src/main/scala/dagr/core/execsystem/TaskStatusReporter.scala
nh13: I'll keep this PR focused for now on the current changes, but definitely this will be up soon: https://github.com/fulcrumgenomics/dagr/issues/74

core/src/main/scala/dagr/core/execsystem/TaskStatusReporter.scala
nh13: bingo

core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
nh13: enclosing node; I like that name better, I will use it for all `parent` parameters.

core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
nh13: It's done, but it was really painful, since `BigInt` was used everywhere, and more importantly, also in `Resource`, which I wanted to keep as `BigInt`.

core/src/main/scala/dagr/core/execsystem/TaskExecutionInfo.scala
nh13, line 45: I will make this change in the next PR.
See: https://github.com/fulcrumgenomics/dagr/issues/76

core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
tfenne: Thanks!

core/src/main/scala/dagr/core/tasksystem/Task.scala
nh13: Done, this will get changed later to track the `TaskExecutionInfo` (see https://github.com/fulcrumgenomics/dagr/issues/76).

nh13:
@tfenne lets chat offline about your last comment, as I think it warrants discussion to get it right.  Other than that, I made most of the changes, and the ones I didn't make are new issues (to make the full re-factor in small chunks).

PR #77, opened 2016-02-11
Some style fixup (e.g. Option.get elimination) in ReflectiveBuilder and friends.

tfenne:
@nh13 my own attempt at ridding ReflectiveBuilder of `Option.get` and `List.head`.

### DIFF
--- a/core/src/main/scala/dagr/core/cmdline/parsing/ClpReflectiveBuilder.scala
+++ b/core/src/main/scala/dagr/core/cmdline/parsing/ClpReflectiveBuilder.scala
@@ -113,7 +113,7 @@ private[parsing] class ClpArgumentLookup(args: ClpArgument*) extends ArgumentLoo
   def names : Set[String] = byName.keySet.toSet // call to set to return an immutable copy
 
   /** Returns the ArgumentDefinition, if one exists, for the provided argument name. */
-  def argFor(argName: String) : Option[ClpArgument] = this.byName.get(argName)
+  def forArg(argName: String) : Option[ClpArgument] = this.byName.get(argName)
 }
 
 /**
--- a/core/src/main/scala/dagr/core/util/ReflectionUtil.scala
+++ b/core/src/main/scala/dagr/core/util/ReflectionUtil.scala
@@ -25,6 +25,8 @@ package dagr.core.util
 
 import java.lang.reflect.Constructor
 
+import dagr.DagrDef._
+
 import scala.annotation.ClassfileAnnotation
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
@@ -168,8 +170,8 @@ object ReflectionUtil {
       val helper = new ReflectiveBuilder[A](clz)
 
       args.foreach { arg =>
-        val name  = arg.children.headOption.map(_.toString).getOrElse(throw new IllegalStateException("Annotation param without name"))
-        val argdef = helper.argumentLookup.fieldFor(name).getOrElse(throw new IllegalStateException("Args must be in lookup"))
+        val name  = arg.children.headOption.map(_.toString) getOrElse unreachable("Annotation param without name")
+        val argdef = helper.argumentLookup.forField(name)   getOrElse unreachable("Args must be in lookup")
         var value = reifyAnnotationParameter(arg.children.tail.head)
 
         if (value.getClass.isArray) {
--- a/core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
+++ b/core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
@@ -3,6 +3,7 @@ package dagr.core.util
 import java.lang.reflect.{Constructor, InvocationTargetException}
 
 import dagr.core.cmdline._
+import dagr.DagrDef._
 
 import scala.collection.mutable
 import scala.reflect.runtime.universe._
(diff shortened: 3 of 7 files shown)


core/src/main/scala/dagr/core/util/ReflectionUtil.scala
```diff
@@ -169,7 +169,7 @@ object ReflectionUtil {
 
       args.foreach { arg =>
         val name  = arg.children.headOption.map(_.toString).getOrElse(throw new IllegalStateException("Annotation param without name"))
-        val argdef = helper.argumentLookup.fieldFor(name).getOrElse(throw new IllegalStateException("Args must be in lookup"))
+        val argdef = helper.argumentLookup.forField(name).getOrElse(throw new IllegalStateException("Args must be in lookup"))
```
nh13: `unreachable`?

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
```diff
@@ -59,41 +60,44 @@ class ReflectiveBuilder[T](val clazz: Class[T]) {
...
-        val paramClass = ReflectionUtil.typeToClass(paramType)
-        val paramUnitType = if (paramType.typeArgs.size == 1) paramType.typeArgs.head else paramType
-        val paramUnitClass = ReflectionUtil.typeToClass(paramUnitType)
-        val paramPrimitiveClass = ReflectionUtil.ifPrimitiveThenWrapper(paramUnitClass)
-
-        val paramTypeName = paramUnitType match {
-          case ref: TypeRef => ref.sym.name.toString
-          case _ => paramType.typeSymbol.asClass.name.decodedName.toString
+    constructor.paramLists match {
+      case Nil | _ :: _ :: _ =>
```
nh13, line 64: can you name the `_` just to make it clearer (I could see them at first).  How about `xs :: rest :: Nil`?

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
```diff
@@ -59,41 +60,44 @@ class ReflectiveBuilder[T](val clazz: Class[T]) {
...
+            case ref: TypeRef => ref.sym.name.toString
+            case _ => paramType.typeSymbol.asClass.name.decodedName.toString
+          }
+
+          // Sort out the default argument if one exists
+          val defaultValue: Option[_] = (companionMembers, instanceMirror) match {
+            case (None, None) => None
+            case (Some(mem), Some(mir)) =>
+              val defarg = mem.member(TermName("$lessinit$greater$default$" + (index + 1)))
+              if (defarg != NoSymbol) Option(mir.reflectMethod(defarg.asMethod)()) else None
```
nh13: perhaps a comment that we should never hit another case, and if we do, a match exception will be thrown.

core/src/main/scala/dagr/core/util/ReflectionUtil.scala
nh13: On second thought, this was a silly idea.

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
```diff
@@ -115,32 +119,23 @@ class ReflectiveBuilder[T](val clazz: Class[T]) {
...
   /** Attempts to build an instance using only default values. Will fail if any argument does not have a default. */
-  def buildDefault() : T = build(this.argumentLookup.ordered.map { a =>
-    a.value match {
-      case Some(v) => v
-      case None => throw new IllegalArgumentException(s"No default value for '${a.name}'")
-    }
-  })
+  def buildDefault() : T = {
+    build(this.argumentLookup.ordered.map(a => a.value getOrElse { throw new IllegalStateException(s" Missing ${a.name}") }))
+  }
```
nh13: extra space in front of `Missing`.

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
```diff
@@ -169,7 +164,7 @@ class ArgumentLookup[ArgType <: Argument](args: ArgType*) {
   def ordered : Seq[ArgType] = argumentDefinitions.toList.sortBy(_.index)
 
   /** Returns the ArgumentDefinition, if one exists, for the provided field name. */
-  def fieldFor(fieldName: String) : Option[ArgType] = this.byFieldName.get(fieldName)
+  def forField(fieldName: String) : Option[ArgType] = this.byFieldName.get(fieldName)
```
nh13, line 168: I thought we abandoned `for...(` in favor of `...For(`.

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
```diff
@@ -221,8 +216,8 @@ class Argument(val declaringClass: Class[_],
 
   /** true if the field has been set, either by a default value or by the user */
   def hasValue: Boolean = _value match {
-    case None               => false
-    case _ if isCollection  => !ReflectionUtil.isEmptyCollection(this._value.get)
-    case _                  => true
+    case None                    => false
+    case Some(c) if isCollection => !ReflectionUtil.isEmptyCollection(c)
+    case _                       => true
```
nh13, line 222: classic tfenne

core/src/main/scala/dagr/core/cmdline/parsing/ClpReflectiveBuilder.scala
```diff
@@ -113,7 +113,7 @@ private[parsing] class ClpArgumentLookup(args: ClpArgument*) extends ArgumentLoo
   def names : Set[String] = byName.keySet.toSet // call to set to return an immutable copy
 
   /** Returns the ArgumentDefinition, if one exists, for the provided argument name. */
-  def argFor(argName: String) : Option[ClpArgument] = this.byName.get(argName)
+  def forArg(argName: String) : Option[ClpArgument] = this.byName.get(argName)
```
nh13, line 116: ditto about the `for...(`/`...For(` debate.

nh13:
:+1: fix up what you want, and merge.

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
tfenne, line 64: Unfortunately scala won't let you name things when doing these multiple matches, since e.g. in this case if it's matches Nil, any names in the second case would be invalid on the right.  I agree it's suboptimal.  The other option it to put this as the last match with just `case _`, but I don't like having a one-line case after a large case body.

core/src/main/scala/dagr/core/util/ReflectiveBuilder.scala
tfenne, line 168: I come back to the idea that it should read well in a sentence, e.g.:

``` scala
val fieldName = "foo"
val arg = argumentLookup.fieldFor(fieldName)  // tells me this should return a Field object
val arg = argumentLookup.forField(fieldName) // tells me it should look something (an arg?) up for a field
val arg = argumentLookup.argFor(fieldName) // tells me it should look up at arg for this field name
```

The last two read sensibly to me, the first one doesn't.  I would characterize as:
1. Use `forP` when you want the _parameter type_ made clear - `P` is the parameter type
2. Use `rFor` when you want the _return type_ made clear - `r` is the thing being returned
3. Use `rForP` when you want to make both the return type and parameter type clearer

tfenne:
@nh13 I've fixed up what I can, and commented on the rest.  I'll leave it here for a while for you to comment, and will merge later tonight or tomorrow morning if you don't have more to say.

nh13:
:+1:

PR #79, opened 2016-02-11
moving the reference to TaskExecutionInfo directly into task

nh13:
See https://github.com/fulcrumgenomics/dagr/issues/76

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
@@ -24,21 +24,14 @@
 package dagr.core.execsystem
 
 import java.nio.file.Path
-import java.sql.Timestamp
 import java.time.Instant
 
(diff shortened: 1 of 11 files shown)


core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
```diff
@@ -99,13 +93,29 @@ trait TaskTracker extends TaskManagerLike with LazyLogging {
         checkForCycles(task = task)
 
         // set the task id
-        val id = nextId
-        task._id = Some(id)
-        nextId += 1
+        val id = yieldAndThen(nextId)(nextId += 1)
```
tfenne: I tend to prefer `yieldAndThen(nextId) {nextId += 1}` braces for the second param list make it look more like a code block.  Up to you though.

tfenne:
:shipit:

Issue #75, opened 2016-02-11
Move TaskRunner interface to a trait


nh13:
... to allow other implementations of `TaskRunner` in the future, and a clearer API.

nh13:
merged

PR #81, opened 2016-02-11
extracting methods for task execution runners

nh13:
Solves https://github.com/fulcrumgenomics/dagr/issues/75


core/src/main/scala/dagr/core/execsystem/TaskExecutionRunner.scala
```diff
@@ -32,6 +32,42 @@ import dagr.core.util.LazyLogging
 
 import scala.collection.mutable
 
+/** A generic template for task execution managers. */
+private[core] trait TaskExecutionRunnerLike {
```
tfenne: I don't love the `FooLike` names.  How about `TaskExecutionRunnerApi`?

tfenne:
One quibble on naming (what else!?), and otherwise looks good.

PR #83, opened 2016-02-11
making a commons package and moving all command line parsing to sopt

nh13:


nh13:
@tfenne just to keep you aware, please see the latest commit message.  If you want to see it in action, try adding the following example the branch: https://gist.github.com/nh13/da11608898eaab1de844

tfenne:
I was thinking about this some, and want to bring up a couple of things that I'd like to change as we extract all this into `sopt`.  This is all centered around annotation support in scala being weak, and generally the preference in the community seeming to be to do things without annotations where possible.  I'm not implying you should do all this - but that it needs to be done and hence maybe we work together on a branch to make this happen?

### No more `@CLP`

I'd like to get rid of the `@CLP` annotation altogether.  I think we could do this by having a `Program` trait in `sopt` that defines a single method `programProperties` or similar (better names needed) that returns an instance of a case class defined in `sopt` like:

``` scala
trait ProgramProperties {
  /* Returns the program group within which the program should be displayed. */
  def group: ProgramGroup
  /* Returns a long description of the program that is displayed with usage. */
  def description: String
  /* Should this program be hidden in command line help. */
  def hidden: Boolean
  /* Gets the short/one-line description displayed when listing programs. */
  def shortDescription: String = StringUtil.firstLine(description)
}

/** An actual implementation. */
case class ProgramProps(group: ProgramGroup, usage: String, hidden: Boolean) extends ProgramProperties
```

I think this has a ton of advantages, including: a) none of the restrictions around annotations, b) users of the framework can implement their own versions of ProgramProperties that generate usage strings in ways we haven't envisioned, c) you can't forget to implement the method that returns this object but you can forget to add an annotation.

### Making group more scala-esque

I also think that `ProgramGroup` becomes just a regular trait also that is extended with objects, e.g.:

``` scala
trait ProgramGroup extends Ordered[ProgramGroup] {
    def name: String = getClass.getName.split("\\$").last
    def description: String
    def rank: Int = 1000
    final def compare(that: ProgramGroup) = this.rank - that.rank match {
      case 0 => this.name.compareTo(that.name)
      case n => n
    }
}

object TimsPrograms extends ProgramGroup {
    val description = "Tim's programs - hands off!"
}
```

Much nicer than a) having to declare a `Class` for each group and b) having `sopt` have to instantiate the class just to get static properties from it.

### Get rid of `@CLPConstructor`

Poking around in the scala reflection API it does seem like there is a way to identify and retrieve the _primary_ constructor for a class.  I think we set things up so that we support:
1. The primary constructor if it is public
2. A non-primary constructor if the class has only a single public constructor
3. Nothing else

### Make `@Arg` -> `@arg` to fit in with scala naming convention

This one is pretty self explanatory.

nh13:
@tfenne this is a big one, and have been a lot of changes, but the major ones are
- creating a `commons` package, which now contains some of the reflection, IO, and util code.
- `sopt` contains not only the option parsing code but now also the command line parsing code.
- many licenses were intellij stubs
- this also involved moving a lot of tests in `core` to `commons` or `sopt`, which was a PITA. 

I think the `sopt.cmdline` package and classes could still use some more work, but at least they are in there final home now.

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -236,13 +257,134 @@ object ReflectionUtil {
...
-  /** Fetches the TypeSymbol (a Symbol representing a type, not a Type) for a given class. */
-  def classToTypeSymbol(clazz: Class[_]): TypeSymbol = {
-    mirror.classSymbol(clazz)
-  }
-
   /** Gets the annotation of type `T` of the class `clazz` */
   def findJavaAnnotation[T <: java.lang.annotation.Annotation](clazz: Class[_], annotationClazz: Class[T]): Option[T] = {
     Option(clazz.getAnnotation(annotationClazz))
   }
+
```
nh13, line 255: these methods were moved from elsewhere

commons/src/main/scala/dagr/commons/util/StringUtil.scala
```diff
@@ -39,7 +42,7 @@ object StringUtil {
     s.split(" ").foldLeft(Array(""))( (out, in) => {
       if ((out.last + " " + in).trim.length > maxLength) out :+ in
       else out.updated(out.length - 1, out.last + " " + in)
-    }).mkString("n").trim
+    }).mkString("\n").trim
```
nh13, line 44: was a bug!

commons/src/test/java/dagr/commons/reflect/BadEnum.java
```diff
@@ -1,7 +1,7 @@
 /*
```
nh13, line 1: I don't think git got the move correct on this one!

commons/src/test/scala/dagr/commons/reflect/ReflectionUtilTest.scala
```diff
@@ -175,6 +177,135 @@ class ReflectionUtilTest extends UnitSpec {
     an[IllegalArgumentException] should be thrownBy ReflectionUtil.isEmptyCollection(value=2)
   }
 
+  "ReflectionUtil.ReflectionUtil.constructFromString" should "construct an Int from a string for an Int field" in {
```
nh13: tests moved from elsewhere to cover methods move to `ReflectionUtil`.

commons/src/test/scala/dagr/commons/util/BiMapTest.scala
```diff
@@ -0,0 +1,76 @@
+/*
```
nh13, line 1: coverage ftw

commons/src/test/scala/dagr/commons/util/LoggerTest.scala
```diff
@@ -0,0 +1,126 @@
+/*
```
nh13, line 1: new testing class

nh13, line 1: We do use it in tests, but in other packages, so it doesn't count towards coverage in this package!

README.md
```diff
@@ -29,7 +29,7 @@ We are currently in alpha development; please see the [Roadmap](#roadmap).
 
 There are many toolkits available for creating and executing pipelines of dependent jobs; dagr does not aim to be all things to all people but to make certain types of pipelines easier and more pleasurable to write.  It is specifically focused on:
 
-* Writing pipeline that are concise, legible and type-safe
+* Writing pipeline that are concise, legible, and type-safe
```
tfenne: pipelines?

commons/src/main/scala/dagr/commons/CommonsDef.scala
```diff
@@ -62,4 +63,35 @@ object DagrDef {
 
   /** The type of identifier used to uniquely identify tasks tracked by the execution system. */
   type TaskId = BigInt
+
```
tfenne: I really don't think that all these type defs belong in `commons`.  They're very `dagr` specific (`TaskId`), and the paths are even specific to the genomics related tasks.

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -21,16 +21,32 @@
...
+import dagr.commons.CommonsDef._
 
 import scala.annotation.ClassfileAnnotation
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
 import scala.reflect.runtime.{universe => ru}
+import scala.util.{Try, Success, Failure}
+
+object ReflectionUtilException {
+  def format(msg: String, e: Exception): String = {
```
tfenne: Since this is only used in one place that I can see, why not eliminate the object entirely and have the single line of concatenation be inside the secondary constructor for the exception?

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -21,16 +21,32 @@
...
+    msg + ": " + e.getMessage + "\n" + e.getStackTrace.mkString("\n")
+  }
+}
+
+/** Base class for all exceptions thrown by the reflection util */
+class ReflectionUtilException(msg: String) extends RuntimeException(msg) {
```
tfenne: Maybe just call this `ReflectionException`?

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -225,6 +241,11 @@ object ReflectionUtil {
     enumClass.getDeclaredField(term.name.toString).get(null)
   }
 
+  /** Converts a type tag to a class */
+  def typeTagToClass[T: TypeTag]: Class[T] = {
```
tfenne: Nice!  My personal preference, and you can disagree, is to write methods like this as:

``` scala
def typeTagToClass[T: TypeTag]: Class[T] = typeToClass(typeTag[T].tpe).asInstanceOf[Class[T]]
```

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -236,13 +257,134 @@ object ReflectionUtil {
...
+            value.toList.mkString(", ") + "'."
+        )
+    }
+  }
+
+  /**
+    * Attempts to construct one or more unitType values from Strings, and return them as the resultType. Handles
+    * the creation of and packaging into collections as necessary.
+    */
+  def typedValueFromString(resultType: Class[_], unitType: Class[_], value: String*): Try[Any] = Try {
```
tfenne: Should this be private, since it's basically the guarded implementation of `constructFromString`?

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -236,13 +257,134 @@ object ReflectionUtil {
...
+    */
+  private def plural(x: Int) = if (x > 1) "s" else ""
+
+  // Also used for Boolean Options
+  private[reflect] val EnumOptionDocPrefix: String = "Options: "
+  private[reflect] val EnumOptionDocSuffix: String = "."
+
+  /** Gets all the options for an enumeration.
+    */
+  def enumOptions(clazz: Class[_ <: Enum[_ <: Enum[_]]]): Try[String] = Try {
```
tfenne: How about having this just return a Seq[String] and moving the code to turn it into a single string for display be in the calling code?

commons/src/main/scala/dagr/commons/reflect/ReflectiveBuilder.scala
```diff
@@ -31,8 +54,7 @@ class ReflectiveBuilder[T](val clazz: Class[T]) {
   extractParameterInformation(constructor)
 
   /**
-    * Finds the appropriate constructor to reflect on. Either the single constructor, or when there are multiple
-    * the one annotated with @CLPConstructor, or failing that the first from the list.
+    * Finds the appropriate constructor to reflect on. The constructor must be public.
     */
   private def findConstructor : MethodSymbol = pickConstructor {
     typeSignature.members.toSeq.filter(p => p.isConstructor && p.owner == clazzSymbol && p.isPublic).map(_.asMethod)
```
tfenne, line 60: Commenting here because github won't allow me to comment below.  I think pickConstructor should be changed to look two ways now:
1) if there's a single ctor return it
2) if there's more than one, but one of the returns true from `ctor.isPrimaryConstructor` return that one
3) Else exception

commons/src/main/scala/dagr/commons/reflect/ReflectiveBuilder.scala
```diff
@@ -92,6 +114,7 @@ class ReflectiveBuilder[T](val clazz: Class[T]) {
             case (Some(mem), Some(mir)) =>
               val defarg = mem.member(TermName("$lessinit$greater$default$" + (index + 1)))
               if (defarg != NoSymbol) Option(mir.reflectMethod(defarg.asMethod)()) else None
+            case _ => unreachable("Should not reach here.")
```
tfenne: Can you make the message more specific, so that if it ever does happen it's clearer?  E.g. `unreachable("companionMembers and instanceMirror must both be None or Some")`

commons/src/main/scala/dagr/commons/util/StringUtil.scala
```diff
@@ -21,12 +21,15 @@
...
-package dagr.core.util
 
+package dagr.commons.util
+
+import java.nio.file.{Paths, Path}
 import java.text.{DecimalFormat, NumberFormat}
 
 object StringUtil {
 
+
```
tfenne: blanka

commons/src/main/scala/dagr/commons/util/StringUtil.scala
```diff
@@ -117,6 +120,9 @@ object StringUtil {
     builder.toString()
   }
 
+  /** Resolves a path from a String, and then makes the path absolute. Prefer this to PathUtil.pathTo elsewhere. */
+  def pathTo(first: String, more: String*): Path = Paths.get(first, more:_*).toAbsolutePath.normalize
```
tfenne: Why move this here from `PathUtil`?  I would not think to look for it here.

commons/src/test/scala/dagr/commons/CommonsDefTest.scala
```diff
@@ -21,17 +21,18 @@
...
+import dagr.commons.CommonsDef._
+import dagr.commons.util.UnitSpec
 
 import scala.collection.mutable.ListBuffer
 
 /**
   * Tests for DagrDef
   */
-class DagrDefTest extends UnitSpec {
+class CommonsDefTest extends UnitSpec {
```
tfenne: Need to change the name of the class in the `should` phrases below.

commons/src/test/scala/dagr/commons/io/IoTest.scala
```diff
@@ -95,12 +97,27 @@ class IoTest extends UnitSpec {
   }
 
   it should "throw an exception because the parent directory is not writable" in {
-    val dir = tmpdir(writable=false);
+    val dir = tmpdir(writable=false)
+    an[AssertionError] should be thrownBy Io.assertCanWriteFile(PathUtil.pathTo(dir.toAbsolutePath.toString, "somewhere"))
```
tfenne: the argument can just be `dir.resolve("somewhere")`

commons/src/test/scala/dagr/commons/io/IoTest.scala
```diff
@@ -95,12 +97,27 @@ class IoTest extends UnitSpec {
...
+    val dir = tmpdir(writable=true)
     an[AssertionError] should be thrownBy Io.assertCanWriteFile(dir)
   }
 
+  it should "not throw an exception when given a writable file" in {
+    val file = tmpfile(writable=true)
+    Io.assertCanWriteFile(file)
+  }
+
+  "Io.assertCanWriteFiles" should "not throw an exception when given a writable files" in {
```
tfenne: files -> file

commons/src/test/scala/dagr/commons/io/IoTest.scala
```diff
@@ -132,4 +149,8 @@ class IoTest extends UnitSpec {
     val grandchild = PathUtil.pathTo(dir.toAbsolutePath.toString, "grand/child")
     Io.findFirstExtentParent(grandchild).get.toAbsolutePath.toString shouldBe dir.toAbsolutePath.toString
   }
+
+  it should "return None if no parent was found" in {
+    Io.findFirstExtentParent(PathUtil.pathTo("/")) shouldBe 'empty
```
tfenne, line 154: This is a strange test case - this was a private method that was only ever invoked on paths that's don't exist.  I'm not sure how I feel about calling it on paths that do exist, but because they are the root, their parent doesn't.

commons/src/test/scala/dagr/commons/reflect/ReflectionUtilTest.scala
tfenne: `ReflectionUtil.ReflectionUtil.constructFromString` -> `ReflectionUtil.constructFromString`

commons/src/test/scala/dagr/commons/reflect/ReflectionUtilTest.scala
```diff
@@ -249,8 +384,19 @@ class ReflectUtilAnnotationTest extends UnitSpec {
     ReflectionUtil.findScalaAnnotation[ClassAnn,ClassAnnotated] shouldBe Some(new ClassAnn(c=classOf[Thread]))
   }
 
+  it should "work with Scala Enumerations" in {
+    ReflectionUtil.findScalaAnnotation[Ann,SomeEnum.type].get.getClass shouldBe classOf[Ann]
```
tfenne: I think you could write this as `.map(_.getClass) shouldBe Some(classOf[Ann])`

commons/src/test/scala/dagr/commons/reflect/ReflectionUtilTest.scala
```diff
@@ -249,8 +384,19 @@ class ReflectUtilAnnotationTest extends UnitSpec {
...
+  }
+
   "ReflectionUtil.hasScalaAnnotation" should "return if the symbol is annotated with a scala annotation" in {
     ReflectionUtil.hasScalaAnnotation[ClassAnn,ClassAnnotated] shouldBe true
     ReflectionUtil.hasScalaAnnotation[EnumAnn,ClassAnnotated] shouldBe false
   }
+
+  "ReflectionUtil.findJavaAnnotation" should "return the annotation on a class" in {
+    val value = ReflectionUtil.findJavaAnnotation(clazz=classOf[ClassJavaAnnotation], annotationClazz=classOf[TestJavaAnnotation])
+    value shouldBe 'defined
```
tfenne, line 408: Similarly, these two lines could just be `value.map(_.placeholder()) shouldBe Some("value")`

commons/src/test/scala/dagr/commons/util/BiMapTest.scala
```diff
@@ -0,0 +1,76 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package dagr.commons.util
+
+class BiMapTest extends UnitSpec {
```
tfenne, line 27: Reading your tests makes me realize that `BiMap` has some potential problems.  Since it's a map type there should probably be some tests that assert that the association between keys and values are correctly maintained by the map.

E.g. what should the following test case produce:

``` scala
val map = new BiMap[String, String]()
map.add("key1", "val1")
map.add("key2", "val1")
???
```

I believe the current implementation will yield:

```
forward:
key1 -> val1
key2 -> val1

reverse:
val1 -> key2
```

This is not intuitive behaviour!  So I think either BiMap should define that each value may have only one key (in which case `key1 -> val1` should be evicted when adding `key2 -> val1`) _or_ BiMap has `key->value` and `value->Set[Key]`.

commons/src/test/scala/dagr/commons/util/CaptureSystemStreams.scala
```diff
@@ -21,7 +21,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package dagr.core.util
+package dagr.commons.util
```
tfenne, line 24: Since this class is only used in tests, perhaps it should be in the test source path?

commons/src/test/scala/dagr/commons/util/LoggerTest.scala
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  "Logger.sanitizeSimpleClassName" should "display scala classes in a readbale form" in {
+    Logger.sanitizeSimpleClassName(classOf[LoggerTest].getSimpleName) shouldBe "LoggerTest"
+  }
+
+  "Logger" should "log an exception" in {
+    val logger = new Logger(classOf[LoggerTest])
+    val stream = new ByteArrayOutputStream()
+    val pw = new PrintStream(stream)
```
tfenne: Could be inlined in the line below.

commons/src/test/scala/dagr/commons/util/LoggerTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    logger.out = Some(pw)
+    val ex = new IllegalArgumentException("Arg!")
+    logger.exception(ex)
+    stream.toString should include ("Arg!")
+  }
+
+  it should "log a message or throwable when the logger is at the same level" in {
+    Logger.synchronized {
```
tfenne: Why is this necessary?  Logger itself doesn't synchronize, so only other code blocks which explicitly do this are blocked.

commons/src/test/scala/dagr/commons/util/LoggerTest.scala
```diff
@@ -0,0 +1,126 @@
...
+
+        val logger = new Logger(classOf[LoggerTest])
+        val stream = new ByteArrayOutputStream()
+        val pw = new PrintStream(stream)
+        logger.out = Some(pw)
+        val value = "Arg!"
+        val belowValue = "grA!"
+        val aboveValue = "Gar!"
+        val ex = new IllegalArgumentException("Exception")
+        level match {
```
tfenne: I think this would be clearer and easier to maintain if you didn't match, but just always did fives lines of:

``` scala
logger.debug("ThisIsDebug")
logger.info("ThisIsInfo")
...
```

And then below you can just have:

``` scala
if (level.ordinal() <= Level.Debug.ordinal()) output should include ("ThisIsDebug") else output shouldNot include ("ThisIsDebug")
...
```

commons/src/test/scala/dagr/commons/util/StringUtilTest.scala
```diff
@@ -21,36 +21,48 @@
...
 class StringUtilTest extends UnitSpec {
-  "StringUtil.camelToGnu" should "correctly handle simple multi-word variable names" in {
-    StringUtil.camelToGnu("fooBar") should be ("foo-bar")
-    StringUtil.camelToGnu("oneTwoThreeFourFiveSixSeven") should be ("one-two-three-four-five-six-seven")
-  }
 
-  it should "not modify single word variable names" in {
-    StringUtil.camelToGnu("foo") should be ("foo")
+  "StringUtil.wrapString" should "wrap a string with a prefix and suffix" in {
+    StringUtil.wrapString("pre-", "body", "-suffix") shouldBe "pre-body-suffix"
```
tfenne: Since we also do wrapping in the sense of word-wrap, perhaps this method should be called `StringUtil.enclose()`?

commons/src/test/scala/dagr/commons/util/StringUtilTest.scala
```diff
@@ -21,36 +21,48 @@
...
+  "StringUtil.levenshteinDistance" should "compute the levenshtein distance between two strings" in {
+    StringUtil.levenshteinDistance("abc", "abc", 1, 1, 1, 1) shouldBe 0
+    StringUtil.levenshteinDistance("abc", "def", 0, 0, 0, 0) shouldBe 0
 
-  it should "handle single letter names" in {
-    StringUtil.camelToGnu("A") should be ("a")
-    StringUtil.camelToGnu("b") should be ("b")
+
+    // swap
+    StringUtil.levenshteinDistance("abc", "acb", 1, 0, 0, 0) shouldBe 0 // can be achieved using substitutions at no penalty
```
tfenne: Would you mind adding parameter names for the various penalties to these tests?

commons/src/test/scala/dagr/commons/util/StringUtilTest.scala
```diff
@@ -21,36 +21,48 @@
...
+
+    // insertion
+    StringUtil.levenshteinDistance("abc", "abcd", 0, 0, 1, 0) shouldBe 1 // insertions cannot be achieved with other operators
+    StringUtil.levenshteinDistance("abc", "abcd", 0, 0, 0, 0) shouldBe 0
+    StringUtil.levenshteinDistance("abc", "abcd", 10, 10, 1, 10) shouldBe 1
   }
-  
-  "String.formatElapsedTime" should "format the elapsed time in DD:HH:MM:SS" in {
+
+  "StringUtil.formatElapsedTime" should "format the elapsed time in DD:HH:MM:SS" in {
```
tfenne: Not so much a comment on the test, but could we change the format here to be just hours, minutes and seconds?  I find that as soon as there are four fields I struggle to decide if it's days on the front or millis on the back.

commons/src/test/scala/dagr/commons/util/Tags.scala
```diff
@@ -21,14 +21,10 @@
...
-/**
-  * Integration tests for the command line parsing
-  */
-class CommandLineTest extends UnitSpec {
+import org.scalatest.Tag
 
-  // TOOD
-}
+/** Tags for our tests */
+object LongRunningTest extends Tag("LongRunningTest")
```
tfenne: Could we refactor a little here to:
1. Call this TestTags instead of just Tags
2. Place all (one) of our tag objects inside an `object TestTags` so that they don't clutter the namespace of the package as we add more?

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -107,7 +110,7 @@ class DagrCoreMain(
   private var reportPath  : Option[Path] = None
 
   // Initialize the configuration as early as possible
-  Configuration.initialize(this.config)
+  dagr.sopt.util.Configuration.initialize(this.config)
```
tfenne: Something about this feels wrong.  Firstly, I think perhaps the main `Configuration` trait should be pushed down to commons.  Secondly I don't love that `dagr` is initializating `sopt`'s configuration.  Perhaps the pattern should be that any toolkit/app that uses needs config extends the `Configuration` trait from `commons` to provide it with access to the appropriate config instance?

core/src/main/scala/dagr/core/cmdline/parsing/DagrCommandLineParser.scala
```diff
@@ -79,117 +47,25 @@ private[cmdline] class DagrCommandLineParser(val commandLineName: String, val in
...
-      * 2. Print the dagr-only help message if --help is used before a pipeline name.
-      * 3. Print the version if --version is used before a pipeline name.
-      * 4. Complain if (a) no pipeline name is given, or (b) it is not recognized.
-      * 5. Complain if pipeline options are mis-specified.
-      * 6. Print the dagr-and-pipeline help message if --help is used after a pipeline name.
-      * 7. Print the version if --version is used after a pipeline name.
-      * 8. ... execute dagr ...
-      * */
+    // Load any Dagr scripts and add them to the classpath.
+    val mainBlock = (dagr: DagrCoreMain) => { loadScripts(clp=dagr); Unit }
```
tfenne: why not `def mainBlock(dagr: DagrCorMain): Unit = loadScripts(clp=dagr)`

core/src/main/scala/dagr/core/cmdline/parsing/DagrCommandLineParser.scala
```diff
@@ -79,117 +47,25 @@ private[cmdline] class DagrCommandLineParser(val commandLineName: String, val in
...
-          val classes = findPipelineClasses(packageList, omitSubClassesOf = Seq(mainClass), includeHidden = includeHidden).keys.toSet
-          print(dagrArgParser.usage())
-          print(pipelineListUsage(classes, Configuration.commandLineName))
-          print(unknownPipelineErrorMessage(Configuration.commandLineName))
-          None
-        }
-        else { // Case 4-8
-          val dagr = dagrArgParser.instance.get
+    // the method that should be called once the instance has been successfully created
+    val clpBlock: DagrCoreMain => Pipeline => Unit.type = (dagr: DagrCoreMain) => (pipeline: Pipeline) => { dagr.configure(pipeline); Unit }
```
tfenne: Again, why not a def?

core/src/main/scala/dagr/core/cmdline/parsing/DagrCommandLineParser.scala
```diff
@@ -25,53 +25,21 @@ package dagr.core.cmdline.parsing
...
-  /** The maximum line lengths for pipeline descriptions */
-  val MaximumLineLength = 80
-
-  /** Similarity floor for matching in printUnknown **/
-  val HelpSimilarityFloor: Int = 7
-  val MinimumSubstringLength: Int = 5
 }
 
 /** Methods for parsing among top level command line tasks */
 private[cmdline] class DagrCommandLineParser(val commandLineName: String, val includeHidden: Boolean = false) {
```
tfenne: Now that this is _so_ much simpler, and it's only real use is from DagrMain, I think it would be really nice to merge those two classes.

core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -104,124 +51,19 @@ private[core] object Configuration extends ConfigurationLike {
   * that is then mixed into the Configuration object above, and used to create a concrete
   * Configuration trait below.
   */
-private[config] trait ConfigurationLike extends LazyLogging {
-  private[config] def config : Config
+private[config] trait ConfigurationLike extends LazyLogging with dagr.sopt.util.ConfigurationLike {
```
tfenne: I really don't see the point of this separate `ConfigurationLike` trait, and then the separate `Configuration` trait below - I think it makes everything more confusing.  Why not just have `Configuration` traits both here and in `sopt`, and have this `Configuration` trait extend `sopt.Configuration`?  I see no added value in this separation.

core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -104,124 +51,19 @@ private[core] object Configuration extends ConfigurationLike {
...
 
-  /**
-    * Looks up a single value of a specific type in configuration. If the configuration key
-    * does not exist, an exception is thrown. If the requested type is not supported, an
-    * exception is thrown.
-    */
-  def configure[T : TypeTag](path: String) : T = {
-    Configuration.RequestedKeys += path
+  // route error messages to our logger
+  override protected def error(parts: Any*) : Unit = logger.error(parts:_*)
```
tfenne: When `Configuration` was just a class in `dagr` I think having it log errors made sense.  But if it's getting extracted somewhere more common (`sopt` or `commons`) then I don't think it's really appropriate for it to be logging - it should just throw exceptions with well-formatted errors.

sopt/src/main/java/dagr/sopt/cmdline/ClpAnnotation.java
```diff
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2015 Fulcrum Genomics LLC
+ * Copyright (c) $year Fulcrum Genomics
```
tfenne: I don't think we should have `$year` here, we should have the actual year this file was written.
Also, in some places you've added LLC and in some places you've taken it away.

sopt/src/main/scala/dagr/sopt/cmdline/ClpArgumentDefinitionPrinting.scala
```diff
@@ -98,7 +101,7 @@ object ClpArgumentDefinitionPrinting {
     val label = new StringBuilder()
     if (shortName.nonEmpty) label.append("-" + shortName + " " + shortType + ", ")
     label.append("--" + name + longType)
-    stringBuilder.append(wrapString(KGRN, label.toString(), KNRM))
+    stringBuilder.append(StringUtil.wrapString(KGRN, label.toString(), KNRM))
```
tfenne: Just thinking out loud - perhaps the terminal codes could become objects that extend a trait like this:

``` scala
trait TermCode {
    val code: String;
    def apply(String s): String = code + s + "\u001B[0m"
}

case object KGRN(override val code = ""\u001B[32m"") extends TermCode

// usage then looks like
builder.append(KGRN("my green text"))
```

sopt/src/main/scala/dagr/sopt/cmdline/ClpGroup.scala
```diff
@@ -21,18 +21,23 @@
...
+
+package dagr.sopt.cmdline
 
 /**
- * Trait for groups of CommandLinePrograms.
- */
-trait PipelineGroup extends Ordered[PipelineGroup] {
+  * Trait for groups of CommandLinePrograms.
+  */
+trait ClpGroup extends Ordered[ClpGroup] {
```
tfenne: Could we still add `rank` here with a default value that's high enough that other classes can easily go above or below it, and then use that as the first order in `compare`?

sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
```diff
@@ -21,29 +21,33 @@
...
 import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 import scala.reflect.runtime.{universe => ru}
+import scala.util.{Failure, Success}
 
 /**
-  * Extends [[ReflectiveBuilder]] to add support for the arguments annotated with the [[Arg]] annotation
-  * and to use more specific logic to find the appropriate constructor to use to instantiate a [[CLP]] pipeline.
+  * Extends [[ReflectiveBuilder]] to add support for the arguments annotated with the [[arg]] annotation
+  * and to use more specific logic to find the appropriate constructor to use to instantiate a command line
```
tfenne: extra `a` towards the end of the sentence.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineException.scala
```diff
@@ -21,65 +21,66 @@
...
+
+object CommandLineException {
+  def format(msg: String, e: Exception): String = {
     msg + ": " + e.getMessage + "\n" + e.getStackTrace.mkString("\n")
   }
 }
 
 /** Base class for all exceptions thrown by the command line parsing */
-private[core] class CommandLineException(msg: String) extends RuntimeException(msg) {
+class CommandLineException(msg: String) extends RuntimeException(msg) {
```
tfenne: This feels a bit cumbersome.  Why not make CommandLineException a bit more flexible:

``` scala
class CommandLineException(msg: String, e: Option[Exception])
  extends RuntimeException(e.map(x => msg + ": " + x.getMessage + "\n" + x.getStackTrace.mkString("\n")) getOrElse msg)
```

Then each of the children can become:

``` scala
class FooException(msg:String, e: Option[Exception]) extends CommandLineException(msg, e)
```

tfenne: Or actually, why not just chain to `RuntimeException`'s constructor that takes a prior?

``` scala
class CommandLineException(msg: String, e: Option[Exception]) extends RuntimeException(msg, e getOrElse null)
```

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+  /** Lengths for names and descriptions on the command line */
+  val ClpGroupNameColumnLength = 48
+  val ClpGroupDescriptionColumnLength = 45
+  val ClpNameColumnLength = 45
+
+  /** The maximum line lengths for tool descriptions */
+  val MaximumLineLength = 80
+
+  /** Error messages */
+  val AvailableClps = "Available Commands:"
```
tfenne: If we're going to let `sopt` depend on configuration, perhaps we should consider making all these strings configurable?  
hrm

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+  */
+class CommandLineParser[Clp](val commandLineName: String)
+                            (implicit classTag: ClassTag[Clp], typeTag: TypeTag[Clp])
+  extends CommandLineParserStrings {
+
+  private type ClpClass = Class[_ <: Clp]
+  private type ClpGroupClass = Class[_ <: ClpGroup]
+
+  /**
+    * Finds the class of a command line program using the first argument given.
```
tfenne: "first argument given" -> first element in the args array

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+    * @param classes the set of available command line program classes
+    * @param clpName the name of the command line program.
+    * @return the error string.
+    */
+  private def unknownClpError(classes: Set[ClpClass], clpName: String) : String = {
+    val builder = new StringBuilder
+    val distances: mutable.Map[ClpClass, Integer] = new mutable.HashMap[ClpClass, Integer]
+    var bestDistance: Int = Integer.MAX_VALUE
+    var bestN: Int = 0
+    classes.foreach{ clazz =>
```
tfenne: Could some of this be factored out into a `findSimilar(target:String, options:Traversable[String): Seq[String]`?  It feels like we have similar code in several places now.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+
+    if (withPreamble) {
+      builder.append(s"${standardClpUsagePreamble()}\n")
+      builder.append(wrapString(KRED, s"$version\n", KNRM))
+    }
+
+    builder.append(wrapString(KBLDRED, s"$AvailableClps\n", KNRM))
+    val clpGroupClassToClpGroupInstance: mutable.Map[ClpGroupClass, ClpGroup] = new mutable.HashMap[ClpGroupClass, ClpGroup]
+    val clpsByGroup: java.util.Map[ClpGroup, ListBuffer[ClpClass]] = new java.util.TreeMap[ClpGroup, ListBuffer[ClpClass]]()
+    val clpsToProperty: mutable.Map[ClpClass, clp] = new mutable.HashMap[ClpClass, clp]
```
tfenne: clpsToClpAnnotation?

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+
+    classes.foreach { clazz =>
+      findClpAnnotation(clazz) match {
+        case None => throw new BadAnnotationException(s"The class '${clazz.getSimpleName}' is missing the required @clp annotation.")
+        case Some(clp) =>
+          clpsToProperty.put(clazz, clp)
+          val clpGroup: ClpGroup = clpGroupClassToClpGroupInstance.get(clp.group) match {
```
tfenne: Wouldn't it be simpler to work with class objects for groups instead of instances, up until you need the properties for printing down below?

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+    *       [[CommandLineProgramParser]] (Clp parser) for the given command line program, and allows the Clp parser to
+    *       parse the rest of the args.
+    *   (3) Returns `None` in the case of any parsing error, or an instance of the command line program constructed
+    *       according to the parsed arguments.
+    *
+    * @param args              the command line arguments to parse.
+    * @param packageList       the list of packages to search for classes that extend [[Clp]].
+    * @param omitSubClassesOf  individual classes to omit from including on the command line.
+    * @param includeHidden     true to include classes whose [[clp]] annotation's hidden values is set to false.
+    * @param clpBlock          a block of code to execute after a [[clp]] }instance has been created successfully.
```
tfenne: This needs a better name.  Perhaps `afterBuilding` or something like that?  Also technically it's a function not a block ;)

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+    * @return                  a [[clp]] instance initialized with the command line arguments, or [[None]] if
+    *                          unsuccessful.
+    */
+  def parseClp(args: Array[String],
+               packageList: List[String],
+               omitSubClassesOf: Iterable[Class[_]] = Nil,
+               includeHidden: Boolean = false,
+               clpBlock: Clp => Unit.type = c => Unit): Option[Clp] = {
+    parseClpHelper(
```
tfenne: Why not just make parseClpHelper() public, and move the extended documentation there?  It already defaults `extraUsage` to None, so there's really no difference between the two methods.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+        printExtraUsage(None)
+        print(clpListUsage(classToClpAnnotation.keySet, commandLineName, withPreamble=false))
+        print(wrapError(error))
+        None
+      case Left(clazz) =>
+        /////////////////////////////////////////////////////////////////////////
+        // Parse the arguments for the Command Line Program class
+        /////////////////////////////////////////////////////////////////////////
+        val clpParser = new CommandLineProgramParser(clazz)
+        clpParser.parseAndBuild(errorMessageBuilder, args.drop(1)) match {
```
tfenne: What do you think about:
1. Making ParseResult a sealed trait
2. Making the values case classes that extend ParseResult
3. Making failure be `case class Failure(message: String)`
4. Having `CommandLineProgramParser` return `Failure(message)` on failures and not take in the errorMessageBuilder like it does now?

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+                printExtraUsage(clazz=Some(clazz))
+                print(clpParser.usage(withPreamble=false))
+                ex.messages.foreach(msg => print(wrapError(msg)))
+                None
+            }
+        }
+    }
+  }
+
+  /** Method for parsing a main classe's command line arguments as well as a specific command line program's arguments.
```
tfenne: `classe's` -> `class'`

tfenne: I think we should use `command` and `sub-command` to differentiate between the top level and the subordinate programs since that's what every other arg/opt parser calls them.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+    * @param mainBlock         a block of code to execute after a [[MainClass]] (main) instance has been created successfully.
+    * @param clpBlock          a block of code to execute after a [[clp]] instance has been created successfully.
+    * @return                  a [[clp]] instance initialized with the command line arguments, or [[None]] if
+    *                          unsuccessful.
+    */
+  def parseMainAndClp[MainClass](args: Array[String],
+                                 packageList: List[String],
+                                 omitSubClassesOf: Iterable[Class[_]] = Nil,
+                                 includeHidden: Boolean = false,
+                                 mainBlock: MainClass => Unit.type = (t: MainClass) => Unit,
```
tfenne: `afterCommandBuilt`?

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+                                 clpBlock: MainClass => Clp => Unit.type = (t: MainClass) => (c: Clp) => Unit)
```
tfenne: `afterSubcommandBuild`?  
Also, I don't think it's useful (and is likely confusing) to take the command and sub-command in separately argument lists.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+      override def genericClpNameOnCommandLine: String = thisParser.genericClpNameOnCommandLine
+    }
+
+    val print : (String => Unit) = System.err.println
+    val errorMessageBuilder: mutable.StringBuilder = new StringBuilder()
+    val (mainClassArgs, clpArgs) = splitArgs(args, packageList, Seq(mainClazz), errorMessageBuilder, includeHidden = includeHidden)
+
+    /** A brief developer note about the precedence of printing error messages and usages.
+      *
+      * 1. Complain if dagr options are mis-specified.
```
tfenne: dagr -> Main/command

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+      * 3. Print the version if --version is used before a clp name.
+      * 4. Complain if (a) no clp name is given, or (b) it is not recognized.
+      * 5. Complain if clp options are mis-specified.
+      * 6. Print the dagr-and-clp help message if --help is used after a clp name.
+      * 7. Print the version if --version is used after a clp name.
+      * 8. ... execute dagr ...
+      * */
+
+    /////////////////////////////////////////////////////////////////////////
+    // Try parsing and building DagrMain and handle the outcomes
```
tfenne: DagrMain -> main/command class.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -0,0 +1,527 @@
...
+        }
+    }
+  }
+
+  /** Splits the given args into two Arrays, first splitting based on a "--", and if not found,
+    * searching for a program name.  If the user specified a Dagr script, a "--" must always be used.
+    * The "--" arg will not be in either returned array.
+    */
+  private[cmdline] def splitArgs(args: Array[String],
+                                                       packageList: List[String],
```
tfenne: indentation

tfenne:
@nh13 I'm about 3/4 of my way through the review and have to run out.  _Please_ don't push to this branch until I've had a chance to finish my review later today.  Thanks!

sopt/src/main/scala/dagr/sopt/cmdline/SpecialArgumentsCollection.scala
```diff
@@ -21,33 +21,22 @@
...
  */
-package dagr.core.cmdline.parsing
+package dagr.sopt.cmdline
 
-import dagr.core.cmdline.Arg
+import dagr.sopt._
 
-private[parsing] object SpecialArgumentsCollection {
+private[cmdline] object SpecialArgumentsCollection {
   val HELP_FULLNAME: String = "help"
```
tfenne: There's aren't used below - we should kill them or use them.

sopt/src/main/scala/dagr/sopt/cmdline/SpecialArgumentsCollection.scala
```diff
@@ -21,33 +21,22 @@
...
-  *   Args: "--aab" "s" "--b" "-c" "-d" "wer we" "-c"
-  *
-  *   Example 2:
-  *   File: "--aab s --b -c -d wer\nwe -c"
-  *   Args: "--aab" "s" "--b" "-c" "-d" "wer", "we" "-c"
   */
-private[parsing] final case class SpecialArgumentsCollection(
-  @Arg(flag = "h", name = "help", doc = "Display the help message.", special = true)
+private[cmdline] final case class SpecialArgumentsCollection(
+  @arg(flag = "h", name = "help", doc = "Display the help message.", special = true)
```
tfenne: Not new to this PR, but I still don't love the way the special arguments work.  Would be nice to find a tidier way to do this.

sopt/src/main/scala/dagr/sopt/util/Configuration.scala
```diff
@@ -0,0 +1,248 @@
...
+        .withFallback(ConfigFactory.defaultApplication())
+        .withFallback(ConfigFactory.defaultReference())
+        .resolve()
+  }
+
+  /** Allows initialization with a custom configuration. */
+  protected def initialize(customConfig: Config): Unit = this._config = customConfig
+
+  /** Returns a sorted set of all keys that have been requested up to this point in time. */
+  private[util] def requestedKeys: SortedSet[String] = {
```
tfenne: This should probably be public, otherwise it's not of much use.  The idea here was that dagr would track all the config keys used and be able to report the full set requested, e.g. during simulation.

sopt/src/main/scala/dagr/sopt/util/Configuration.scala
```diff
@@ -0,0 +1,248 @@
...
+
+  private val name = sanitizeSimpleClassName(getClass.getSimpleName)
+
+  /** Removes various characters from the simple class name, for scala class names. */
+  private def sanitizeSimpleClassName(className: String): String = {
+    className.replaceFirst("[$].*$", "")
+  }
+
+  /** The method to write any errors, by default writing to standard error. */
+  protected def error(parts: Any*) : Unit = {
```
tfenne: I was thinking more on this ... how about we remove this from this `Configuration` and in Dagr's configuration override `configure[T]()` to call super, trap the exception, log it, then rethrow it?

sopt/src/main/scala/dagr/sopt/util/Configuration.scala
```diff
@@ -0,0 +1,248 @@
...
+  /**
+    * Attempts to determine the path to an executable, first by looking it up in config,
+    * and if that fails, by attempting to locate it on the system path. If both fail then
+    * an exception is raised.
+    *
+    * @param path the configuration path to look up
+    * @param executable the default name of the executable
+    * @return An absolute path to the executable to use
+    */
+  def configureExecutable(path: String, executable: String) : Path = {
```
tfenne: I think perhaps this and configureExecutableFromBinDirectory belong in dagr's Configuration trait.

sopt/src/main/scala/dagr/sopt/util/ParsingUtil.scala
```diff
@@ -0,0 +1,141 @@
...
+import dagr.sopt.cmdline._
+import dagr.sopt.clp
+
+import scala.collection.JavaConversions._
+import scala.collection.immutable.Map
+import scala.reflect._
+import scala.reflect.runtime.universe._
+
+/** Variables and Methods to support command line parsing */
+object ParsingUtil extends Configuration  {
```
tfenne: Again, not new, but I've never liked that scala makes you say extend for the first trait if you're not extending another class.  How do you feel about doing this instead?:

``` scala
object ParsingUtil extends Object with Configuration
```

I think that makes it clearer that you're "adding" configuration, rather than being an extension of it.

sopt/src/main/scala/dagr/sopt/util/ParsingUtil.scala
```diff
@@ -0,0 +1,141 @@
...
+    * @return a mapping between sub-classes of [[T]] and their [[clp]] annotation.
+    */
+  private def classToPropertyMapFromSourceClasses[T](srcClasses: Traversable[Class[_]],
+                                                     omitSubClassesOf: Iterable[Class[_]] = Nil,
+                                                     includeHidden: Boolean = false)
+  : Map[Class[_ <: T], clp] = {
+
+    // Filter out interfaces, synthetic, primitive, local, or abstract classes.
+    def keepClass(clazz: Class[_]): Boolean = {
+      !clazz.isInterface && !clazz.isSynthetic && !clazz.isPrimitive && !clazz.isLocalClass && !Modifier.isAbstract(clazz.getModifiers)
```
tfenne, line 62: Should we also check for `public`?

sopt/src/main/scala/dagr/sopt/util/ParsingUtil.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  /** Gets a mapping between sub-classes of [[T]] and their [[clp]] annotation.
+    *
+    * @param srcClasses the source classes from which to build the map.
+    * @param omitSubClassesOf a list of omit subclasses to omit from the returned map.
+    * @param includeHidden include classes whose [[clp]] annotation has `hidden` set to true.
+    * @tparam T the super class.
+    * @return a mapping between sub-classes of [[T]] and their [[clp]] annotation.
+    */
+  private def classToPropertyMapFromSourceClasses[T](srcClasses: Traversable[Class[_]],
```
tfenne: property -> annotation

sopt/src/main/scala/dagr/sopt/util/ParsingUtil.scala
```diff
@@ -0,0 +1,141 @@
...
+    if (nameCollisions.nonEmpty) {
+      throw new CommandLineException(s"Simple class name collision: ${nameCollisions.mkString(", ")}")
+    }
+
+    // Finally, make the map
+    classes.map(clazz => Tuple2(clazz, ReflectionUtil.findJavaAnnotation(clazz, classOf[clp]).get)).toMap
+  }
+
+
+  /** Gets a mapping between sub-classes of [[T]] and their properties.
```
tfenne: The comment is a bit out of date since this is where the actual searching for the classes happens.

sopt/src/test/scala/dagr/sopt/cmdline/ClpReflectiveBuilderTest.scala
```diff
@@ -0,0 +1,118 @@
...
+
+package dagr.sopt.cmdline
+
+import dagr.sopt.arg
+import dagr.commons.util.UnitSpec
+
+/**
+  *
+  */
+private case class IntNoDefault(@arg var v: Int)
```
tfenne: Could the parameters to all these be vals instead of vars?

sopt/src/test/scala/dagr/sopt/cmdline/TestingClp.scala
```diff
@@ -0,0 +1,30 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) $year Fulcrum Genomics
```
tfenne: $year again

tfenne:
Phew - ok, finished going through.  A big improvement!  I have a few general thoughts to add to the litany of line comments:

### StringUtil

I think `StringUtil` should be reserved for utilities that solely manipulate strings (and perhaps other string-like types such as char[]s or byte[]s), and that all functions for converting other types of things to/from Strings should go in separate utilities.  E.g. 
- `pathTo` belongs in `PathUtil`
- `formatElapsedTime` belongs in a `TimeUtil` or `DateTimeUtil` or `ChronoUtil` or something

Other projects we've worked on (e.g. htsjdk and Picard) pushed anything that converted to/from a String into a StringUtil, and then it was very confusing trying to find utility methods for various objects as sometimes they were in StringUtil and sometimes they were in other specific util classes.  Since `commons` is supposed to be our general utility library, let's try and be relatively strict about separating out `*Util` classes and methods by the more specific type they apply to.

This seems simple enough that if you agree, I'd like to fix before merging.

### Configuration

The Configuration split still needs work.  The first question is whether we really want to let `sopt` use configuration at all.  I would lean towards not allowing `sopt` to use configuration, but instead having a clear entry point for people to use, and a way to configure that cleanly in code.  I think this is `CommandLineParser`, so perhaps we could set that up with either a monster constructor full of default values, or a fluent interface (i.e. `New CommandLineParser().withName("dagr").withColors(true)`).

_If_ that makes sense then we still need to decide whether to split config into `commons` and `dagr`.  I'm not sure I feel strongly either way, but there are definitely methods in the `sopt` version which I think belong only in `dagr`.

Lastly, I think the split into `ConfigurationLike` and `Configuration` traits doesn't add functionality or composability, but just confuses the situation.  Sorry.

Again, hopefully if you agree, this is a small enough change to make prior to merging?

### Recursive command parsing

The command line parsing stuff lookes _much_ better already, and I think this might be an improvement for a later refactor as opposed to this PR.  Basically I think that instead of `CommandLineParser` supporting `Clp` and `MainAndClp`, it should support nested `CommandLineParser`s.  I haven't thought this all the way through yet, but in general I'm thinking that we would add a method to CommandLineParser that's something like:

``` scala
def subParser(argSplitter: Array[String] => (Array[String], Array[String]),
              subParser  : CommandLineParser)
```

And that if a sub-parser is set then `CommandLineParser` uses the `argSplitter` function to split it's args array into two, the first of which is for this `CommandLineParser` and the second is for the `subParser`.

E.g. it would be good to support the following (and if this is to become a general library, others will definitely want this kind of thing):

```
dagr --config=foo.conf run --cores=16 MyPipeline --in=foo.txt --out=bar
dagr --config=foo.conf simulate MyPipeline --in=foo.txt --out=bar
dagr --config=foo.conf web --port 8080
```

We'd have to think more about the API, but I'm convinced recursive parsers are the right answer longer term, and would actually make `CommandLineParser` tidier.  But that's probably more than you want to take on for this PR.

commons/src/main/scala/dagr/commons/CommonsDef.scala
nh13: Good catch on the `TaskId`.  I do want to keep the `PathTo*` in here, since I will want o use them for `JeanLuc`.  How about a second def (`BfxPathDef` or something)?

commons/src/main/scala/dagr/commons/util/StringUtil.scala
nh13: It was in `StringUtil` before ;)

commons/src/test/scala/dagr/commons/io/IoTest.scala
nh13, line 154: Perhaps it's too clever.

commons/src/test/scala/dagr/commons/util/CaptureSystemStreams.scala
nh13, line 24: It already is:

```
 find . | grep CaptureSystemStreams.scala$
./commons/src/test/scala/dagr/commons/util/CaptureSystemStreams.scala
```

commons/src/test/scala/dagr/commons/util/LoggerTest.scala
nh13: I'll change it to `this.synchronized` but it modifies the `level` value in the `Logger` singleton.  I'll also add a comment:
`// synchronized since we modify the log level, and we don't want to affect other tests`

commons/src/test/scala/dagr/commons/util/LoggerTest.scala
nh13: sounds good

commons/src/test/scala/dagr/commons/util/StringUtilTest.scala
nh13: I always enjoy the opportunity to use IntelliJ from regex S&R.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
nh13: Deferred.

core/src/main/scala/dagr/core/cmdline/parsing/DagrCommandLineParser.scala
nh13: This caused me to unravel a few things, I don't know why I had `Unit.type` everywhere.

core/src/main/scala/dagr/core/cmdline/parsing/DagrCommandLineParser.scala
nh13: Yup

core/src/main/scala/dagr/core/config/Configuration.scala
nh13: I think it was a hold over from when the `Configuration` object was private.  I'll clean it up.

sopt/src/main/java/dagr/sopt/cmdline/ClpAnnotation.java
nh13: I found a bunch of classes with `$year` and some of them got fixed, and I'll blame IntelliJ for breaking ones like this.

sopt/src/main/scala/dagr/sopt/cmdline/ClpArgumentDefinitionPrinting.scala
nh13: `case object` and `String s`?  I see what you mean though.

commons/src/main/scala/dagr/commons/util/StringUtil.scala
tfenne: I was all ready to fall on my sword and take the blame for this - but then I looked and it's not in `StringUtil` on master...

commons/src/main/scala/dagr/commons/CommonsDef.scala
tfenne: I understand - but I still feel weird about it.  My thought was that commons was to be generally re-usable scala utilities, not just a place to deposit anything we wanted to use across two projects.

commons/src/test/scala/dagr/commons/io/IoTest.scala
tfenne, line 154: My point here, now that I think about it more, is that this test is testing under-specified behaviour that is a consequence of how the function is implemented.  Tests should test specified behaviour, and if things are under-specified we should fix that first, vs. creating tests that assert that the current, under-specified behaviour, is maintained.

In this case the method is effectively private - it's only `private[util]` so that it can be tested, and is never called with a path that exists.

commons/src/test/scala/dagr/commons/util/CaptureSystemStreams.scala
tfenne, line 24: alrighty

sopt/src/main/scala/dagr/sopt/cmdline/ClpArgumentDefinitionPrinting.scala
tfenne: Doh - when I'm writing snippets like this I still sometimes forget and use Java syntax :confounded: 
I did mean `case object` though, although I guess making them `case` objects doesn't really add much over being just `objects` in this case.

nh13: I understand now.

sopt/src/main/scala/dagr/sopt/cmdline/ClpGroup.scala
```diff
@@ -21,18 +21,23 @@
...
-
   /** Order groups by name. */
-  override def compare(that: PipelineGroup): Int = this.name.compareTo(that.name)
+  override def compare(that: ClpGroup): Int = this.name.compareTo(that.name)
 }
 
+/** The program group for the command line programs. */
+class Clps extends ClpGroup {
+  val name: String = "Clps"
+  val description: String = "Various command line programs programs."
```
nh13: doh doh

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineException.scala
nh13: It's really nice to get rid of some of the holdover stuff like this that's left from porting the command line parsing from other Java projects.

nh13: Also, today I learned `e getOrElse null` can be written as `e.orNull`.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
nh13: I was thinking that someone could just mix-in their own trait that extends `CommandLineParserStrings` and overrides accordingly.

nh13: zcsh2

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
nh13: :+1:

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineException.scala
tfenne: I didn’t know that.  That’s super-handy!

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
nh13: Future improvement.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
tfenne: I see.  I think that works.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
nh13: Per the latter, I refer you to `yieldAndThen` which someone wrote:

```

  /**
    * Construct to capture a value, execute some code, and then returned the captured value. Allows
    * code like:
    *   `val x = foo; foo +=1; return x`
    * to be replaced with
    *   `yieldAndThen(foo) {foo +=1}`
    * @param it the value to be returned/yielded
    * @param block a block of code to be evaluated
    * @tparam A the type of thing to be returned (usually inferred)
    * @return it
    */
  def yieldAndThen[A](it: => A)(block: => Unit): A = {
    val retval : A = it
    block
    retval
  }
```

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
nh13: Great idea of course, but so much work!  Will do it.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
tfenne: Imma gonna win this argument :stuck_out_tongue:  `clpBlock` here is actually a function type because it's `Clp => Unit` - i.e. it has a parameter list and a return type.  `block` in `yieldAndThen` is a call-by-name parameter, which is (frustratingly) not a function type in scala, hence the need to play silly tricks like this in EitherTask:
`def apply(left: Task, right: Task, choice: => Choice): EitherTask = new EitherTask(left, right, () => choice)`

Ahh, technically right, the best kind of right :wink:

nh13: image

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
nh13: I settled on keeping the enumeration (since empty case classes are deprecated) but instead using a `Try` to indicate success or failure.  I also added a method to the `CommandLineProgramParser` object to perform all the usage printing in the case of help, version, or failure.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
nh13: I don't understand the last comment, could you restate it?

sopt/src/main/scala/dagr/sopt/cmdline/SpecialArgumentsCollection.scala
nh13: https://github.com/fulcrumgenomics/dagr/issues/101

sopt/src/main/scala/dagr/sopt/util/ParsingUtil.scala
nh13: I don't have a problem with either.

sopt/src/main/scala/dagr/sopt/util/ParsingUtil.scala
nh13, line 62: No, folks may want to hide the class from others, but still make it available.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
tfenne: The way `clpBlock` is specified will require invokers of this method to write code like:

``` scala
... clpBlock = (main) => (clp) => { someCode }
```

instead of the following which seems more intuitive:

``` scala
... clpBlock = (main, clp) => { someCode }
```

sopt/src/main/scala/dagr/sopt/util/ParsingUtil.scala
tfenne, line 62: As it stands, that won't work unless we do some hackery to override the access modifier on the class and/or constructor.  We'll get a security violation exception when we try to instantiate it.

nh13:
@tfenne, I made most of your suggested changes. I created a `GenomicsDef` in `commons` so that other tools can access to those typedefs.  I would like your input to decide if that is the path forward (or not).

Furthermore, I have two more things to do before this PR is ready:
1. try to implement recursive command parsing
2. move all Configuration back into core

nh13:
I am going to leave the recursive sub-parsing to another day, as I think the interface needs some thought.  Therefore, this PR is ready for a squash and merge after a final :+1: from @tfenne.

### Recursive parsing

 It would be nice to provide a curried method for folks to use the current implementation of `splitArgs` in `CommandLineParser`:

```
def splitArgs[SubCommand](packageList: List[String],
              omitSubClassesOf: Iterable[Class[_]] = Nil,
              includeHidden: Boolean = false)(args: Array[String]): (Array[String], Array[String])
```

so that someone could just pass in:

```
splitArgs = splitArgs[BaseSubCommandClass](packageList=List("some.package."))
```

and get the default splitting of the class names for free.

Note that we need to know both how to split the args as well as the base class for the sub-commands as it currently stands, and this would continue as we recursive.  I don't know how you would return the instances though, since you might want `Option[DagrCoreMain, Simulate, Pipeline]` or `Option[DagrCoreMain, Run, PipelineTwo]`.  Perhaps we should look to Akka routing to see if there is a better way to specify nested sub-parsers?

tfenne:
@nh13 I'm not going to re-read the code.  I think the decisions you've made sound good - particularly the recursive parsing seemed like too big to bundle into this PR.  Go ahead and squash and merge.

Issue #88, opened 2016-02-12
Dagr should support task args with spaces and special characters


tfenne:
This does not go well:

```
java -Xmx1g -jar target/scala-2.11/dagr-0.1.0-SNAPSHOT.jar \ 
  --config=./resources/application.conf \
  DnaResequencingFromFastqPipeline \
  --fastq1 r1.fq.gz \
  --fastq2 r2.fq.gz \
  --ref /Work/refseq/hs38DH/hs38DH.fa \
  -s 'My Stupid Sample' \
  -l 'Foo*Bar' \
  -p Nope\! \
  --tmp /tmp \
  -o ./breakage
```

It generates a script that contains the following:

```
run () {
  java -Dsamjdk.buffer_size=131072 -XX:GCTimeLimit=50 -XX:GCHeapFreeLimit=10 -Xmx4096m -jar /Work/dgx/packages/picard/picard.jar FastqToSam VALIDATION_STRINGENCY=SILENT CREATE_INDEX=true F1=/Work/dgx/data/test/r1.fq.gz F2=/Work/dgx/data/test/r2.fq.gz O=/dev/stdout SM=My Stupid Sample PL=ILLUMINA PU=Nope! LB=Foo*Bar STRIP_UNPAIRED_MATE_NUMBER=true SO=queryname \
    |  java -Dsamjdk.buffer_size=131072 -XX:GCTimeLimit=50 -XX:GCHeapFreeLimit=10 -Xmx1024m -jar /Work/dgx/packages/picard/picard.jar FifoBuffer VALIDATION_STRINGENCY=SILENT CREATE_INDEX=true BUFFER_SIZE=536870912 \
    |  java -Dsamjdk.buffer_size=131072 -XX:GCTimeLimit=50 -XX:GCHeapFreeLimit=10 -Xmx1024m -jar /Work/dgx/packages/picard/picard.jar MarkIlluminaAdapters VALIDATION_STRINGENCY=SILENT CREATE_INDEX=true I=/dev/stdin O=/tmp/unmapped.4895479567643351291.bam M=/Work/dgx/data/test/breakage/Foo*Bar.adapter_metrics.txt
}
```

To fix this we'll need to do two things:
1. Change the code that builds up the script to either quote and escape every single argument (ugly), or detect which arguments need quoting/escaping and apply it just to those (more work, but a lot less ugly)
[...]

PR #89, opened 2016-02-12
Changed ProcessTask to quote and escape arguments as necessary.

tfenne:
Modified Pipe to override ProcessTask.commandLine intead of ProcessTask.args so that the pipes don't get escaped!

### DIFF
--- a/core/src/test/scala/dagr/core/tasksystem/PipeTest.scala
+++ b/core/src/test/scala/dagr/core/tasksystem/PipeTest.scala
@@ -24,12 +24,16 @@
 package dagr.core.tasksystem
 
 import dagr.core.execsystem.{Cores, Memory, ResourceSet}
-import dagr.core.util.UnitSpec
+import dagr.core.util.{PathUtil, UnitSpec}
 
 /**
   * Tests for the piping together of tasks
   */
 class PipeTest extends UnitSpec {
+  val `|`  = Pipes.PipeString
+  val `>`  = Pipes.RedirectOverwriteString
+  val `>>` = Pipes.RedirectAppendString
+
   // Data types used in piping
   object Types {
     class Text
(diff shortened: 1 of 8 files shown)


tfenne:
@nh13 This is a work in progress for #88.  I need to add more tests, but this appears to work (pipelines and existing tests pass).

nh13:
We discussed this in-person, but I will put this here, lets not add single quotes if there are single quotes at the start and end of the string.  Also, it would be really nice to support `>` and `>>` in PipedTask.

core/src/test/scala/dagr/core/tasksystem/PipeTest.scala
```diff
@@ -54,22 +54,21 @@ class PipeTest extends UnitSpec {
...
 
   it should "handle re-use of parts of pipes" in {
-    val `|` = " " + Pipes.PipeString + " "
+    val `|` = Pipes.PipeString
     val start = Cat("foo.txt") | MakeCsv()
     val pipe1 = start | CsvToTsv()
     val pipe2 = start | Column()
-    pipe1.args.mkString(" ") shouldBe "cat foo.txt" + `|` + "sed s/ +/,/g" + `|` + """tr ',' '\t'"""
-    pipe2.args.mkString(" ") shouldBe "cat foo.txt" + `|` + "sed s/ +/,/g" + `|` + "column -t"
+    pipe1.commandLine shouldBe "cat foo.txt" + `|` + "sed 's/ +/,/g'" + `|` + """tr , '\t'"""
```
nh13, line 72: Change `"sed 's/ +/,/g'"` to `"sed", "'s/ +/,/g'"` here and elsewhere since I don't want to quote an arg if it starts and ends with `'`.

nh13:
I made a few comments on the last two commits, but squash and merge is fair too.

tfenne:
@nh13 I've squashed and pushed in preparation for merge, but weren't you going to submit a PR to this branch for the freebayes task(s) before we merged?

nh13:
@tfenne Wayne Gretzky has something to say about it: https://github.com/fulcrumgenomics/dagr/pull/99

tfenne:
@nh13 I merged your PR and then made some tweaks to get your example to do the right thing.  Take a look at commit 2b368c0. It boils down to three things:
1. Removing some characters from the set of specials - I think I was overzealous in including `[]{}` as without other control structures bash treats those as normal characters.
2. Splitting `"--region {}"` into `"--region", "{}"`
3. Removing the single quotes from your awk command

I didn't make the change to not-single-quote a string if it includes single quotes at the start and end already, mostly because I was/am hoping that you would no longer want that change once you'd tried things out.  I won't squash and merge until you've had a final chance to look and object :wink:

tfenne:
@nh13 I'm merging this as it's starting to block me.  We can always revisit the quoting if necessary.

nh13:
:+1:

PR #95, opened 2016-02-12
Example pipeline for the readme.

tfenne:


pipelines/src/main/scala/dagr/pipelines/ExamplePipeline.scala
```diff
@@ -0,0 +1,62 @@
...
+import dagr.tasks.{DirPath, PathToFasta, PathToFastq, PathToIntervals}
+import htsjdk.samtools.SAMFileHeader.SortOrder
+
+/**
+  * Very simple example pipeline that is used in the readme.md
+  */
+@CLP(description="Example FASTQ to BAM pipeline.", group = classOf[Pipelines])
+class ExamplePipeline
+( @Arg(flag="i", doc="Input fastq.")        val fastq: PathToFastq,
+  @Arg(flag="r", doc="Reference FASTA.")    val ref: PathToFasta,
```
nh13: `FASTA` but `fastq`.  Pick a case.

pipelines/src/main/scala/dagr/pipelines/ExamplePipeline.scala
```diff
@@ -0,0 +1,62 @@
...
+import dagr.core.cmdline.{Arg, CLP, Pipelines}
+import dagr.core.tasksystem.Pipeline
+import dagr.core.util.Io
+import dagr.tasks.bwa.BwaMem
+import dagr.tasks.picard.{CollectHsMetrics, DeleteBam, MarkDuplicates, SortSam}
+import dagr.tasks.{DirPath, PathToFasta, PathToFastq, PathToIntervals}
+import htsjdk.samtools.SAMFileHeader.SortOrder
+
+/**
+  * Very simple example pipeline that is used in the readme.md
```
nh13: `readme` should be upper case.

pipelines/src/main/scala/dagr/pipelines/ExamplePipeline.scala
```diff
@@ -0,0 +1,62 @@
...
+  */
+@CLP(description="Example FASTQ to BAM pipeline.", group = classOf[Pipelines])
+class ExamplePipeline
+( @Arg(flag="i", doc="Input fastq.")        val fastq: PathToFastq,
+  @Arg(flag="r", doc="Reference FASTA.")    val ref: PathToFasta,
+  @Arg(flag="t", doc="Target regions.")     val targets: Option[PathToIntervals],
```
nh13: default to `None`?

nh13:
quibbles => :+1:

nh13:
reminder to self: perhaps we have an `apply` method on `Pipeline` that that doesn't take an `Option[Path]` but instead `Path`?

nh13:
for the example Pipeline, do we want to include the imports?

nh13:
We should have a link to the `Linker` task (when ready) in the section " List of features" in the sub-section "Mechanisms for passing state"

tfenne:
@nh13 I don't think having `Pipeline.apply(Path)` helps, because you have to chain to a constructor, not an apply method.  We could create a secondary constructor that takes `Path` instead of `Option[Path]`, but I worry that if we add more things over time (e.g. failure policy, naming context) we'll end up with dozens of constructors.

PR #97, opened 2016-02-13
Removed Callbacks and replaced with much simpler Linker concept.

tfenne:
@nh13 The actual implementation of `Linker` is tiny.  Most of the changes are from
1. ripping out `Callbacks`
2. converting everything over from Callbacks to Linker
3. tidying up SimplerInJvmTask.apply()

The name `Linker` seems fine to me, but if you have better suggestions let me know.


core/src/main/scala/dagr/core/tasksystem/Linker.scala
```diff
@@ -0,0 +1,55 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package dagr.core.tasksystem
+
+import dagr.core.execsystem.{Memory, Cores}
+
+/**
+  * Object for creating linker tasks that transfer state between other tasks at
+  * runtime.
```
nh13, line 30: <not serious> we need to add in the quote: "With the first link, the chain is forged..., chains us all irrevocably", since TNG references are always appropriate for comments and docs.</not serious>

core/src/main/scala/dagr/core/tasksystem/Linker.scala
```diff
@@ -0,0 +1,55 @@
...
+    *
+    * @param parent a task that produces some information on which the child depends
+    * @param child  a task which needs to receive some information from the parent task
+    * @param link   a method that will receive be invoked with a reference to both tasks
+    *               at the appropriate time during pipeline execution
+    * @tparam Parent the type of the parent task (usually inferred)
+    * @tparam Child  the type of the child task (usually inferred)
+    * @return a task that links the parent and child and will invoke the `link` method
+    */
+  def apply[Parent <: Task, Child <: Task](parent: Parent, child: Child)
```
nh13: Make it 

```
apply[Parent <: Task, Child <: Task](parent: Parent)(child: Child)(link: (Parent, Child) => Unit)
```

and it should have a better chance at being inferred.

core/src/main/scala/dagr/core/tasksystem/Task.scala
```diff
@@ -142,9 +142,9 @@ object Task {
  *
  * Once a task is constructed, it has the following evolution:
  * 1. Any tasks on which it depends are added (see [[==>]]).
- * 2. When all tasks on which it is dependent have completed, callbacks (see [[Callbacks]]
- *    are performed and the then [[getTasks]] method is called to create a set of tasks.
- *    This task becomes dependent on any task that is returned that is not itself.
+ * 2. When all tasks on which it is dependent have completed, the [[getTasks]] method
+ *    is called to create a set of tasks. This task becomes dependent on any task that
+  *    is returned that is not itself.
```
nh13: space

core/src/test/scala/dagr/core/tasksystem/TaskTest.scala
```diff
@@ -78,62 +78,58 @@ class TaskTest extends UnitSpec with LazyLogging {
    * Example #1: a "lazy task" that doesn't have all of its required inputs until
    * a previous task on which it is dependent has getTasks called.
    */
+  object Callbacks {}
```
nh13: `Linkers`

core/src/main/scala/dagr/core/tasksystem/Linker.scala
nh13: I am not in love with `Parent` and `Child`, how about `Src` and `Dest`?

core/src/test/scala/dagr/core/tasksystem/TaskTest.scala
```diff
@@ -78,62 +78,58 @@ class TaskTest extends UnitSpec with LazyLogging {
...
       val firstTask = new FirstTaskNeedsBuilding()
       val lazyTask = new LazyTask(sleepTime = firstTask.sleepTime)
-      Callbacks.connect(lazyTask, firstTask)((l,f) => l.sleepTime = f.sleepTime)
-      // So now lazyTask depends on the number from firstTask being not None.
-      // If we call lazyTask.getTasks, it will call firstTask.getNumber, which will fail,
-      // since fistTask.number is None.
-      lazyTask.invokeCallbacks()
-      an[RuntimeException] should be thrownBy lazyTask.applyResources(ResourceSet.infinite)
+
+      // Linker is purposefully backwards
```
nh13: add to the comment that you _set it up to fail_ not that the Linker is inherently supposed to be used backwards.  I read tests to figure how to use things sometimes, and this would confuse me.

nh13:
@tfenne this looks great, and makes callbacks so much more intuitive and easy to write.  Thank-you!  Incorporate my comments or don't, and feel free to merge when ready.

core/src/main/scala/dagr/core/tasksystem/Linker.scala
tfenne: I'm not sure why you're suggesting putting parent and child in separate argument lists - there's no dependence between their types, and so it doesn't help with anything - it just makes the invocation more awkward.  The function is already in a separate argument list so that it's types can be inferred.

tfenne: How about 'from' and 'to'?  That seems right, you create a link `from` x `to` y.

PR #98, opened 2016-02-13
Refactor EitherTask to use call-by-name instead of var.

tfenne:
There's quite a lot going on in this relatively small commit:
1. Refactor so that choice is no longer a var, but is instead a call-by-name parameter that gets wrapped into a function that is evalulated when the choice needs to be made. This removes the need for a callback/linker to use EitherTask.
2. Removed the Enumeration in favor of a sealed trait and case objects as I find their usage to be nicer.
3. Made the constructor private so that EitherTask.apply() is the only way to make an EitherTask. This is to make the API cleaner since call-by-name parameters cannot be vals, so the companion wraps the call-by-name into a function that is passed to the contructor val.
4. I had to make the Boolean version have a different name (EitherTask.of()) because apply(Task,Task,=>Choice) and apply(Task,Task,=>Boolean) have the same type after erasure and the compiler won't allow that :(

### DIFF
--- a/core/src/main/scala/dagr/core/tasksystem/EitherTask.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/EitherTask.scala
@@ -23,34 +23,42 @@
  */
 package dagr.core.tasksystem
 
-/** Helps determine which task to perform for an [[EitherTask]] */
-object EitherTaskFlag extends Enumeration {
-  type EitherTaskFlag = Value
-  val LEFT, RIGHT = Value
-}
-
 object EitherTask {
+  sealed trait Choice
+  object Left  extends Choice
+  object Right extends Choice
 
-  /** Creates an either task given [[Boolean]] flag indicating which task to perform.
+  /**
+    * Creates an [[EitherTask]] that wraps `choice` into a function that will be evaluated lazily when the
+    * [[EitherTask]] needs to make its choice.
     *
-    * @param left the left task.
-    * @param right the right task
-    * @param doLeft true if the left task is to be executed, false otherwise.
-    * @return an [[EitherTask]] with the correct flag set.
+    * @param left   the left task.
+    * @param right  the right task.
+    * @param choice an expression that returns either Left or Right when evaluated
     */
-  def apply(left: Task, right: Task, doLeft: Boolean): EitherTask = {
-    new EitherTask(left, right, if (doLeft) EitherTaskFlag.LEFT else EitherTaskFlag.RIGHT)
-  }
+  def apply(left: Task, right: Task, choice: => Choice): EitherTask = new EitherTask(left, right, () => choice)
+
+  /**
+    * Creates an [[EitherTask]] that wraps `choice` into a function that will be lazily evaluated when the
+    * [[EitherTask]] needs to make its choice. If `goLeft` evaluates to true the `Left` task will
+    * be returned, else the `Right` task.
+    *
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dagr/core/tasksystem/EitherTask.scala
```diff
@@ -23,24 +23,31 @@
...
-}
-
 object EitherTask {
+  sealed trait Choice
+  object Left  extends Choice
+  object Right extends Choice
+
+  /**
+    * Creates an [[EitherTask]] that wraps `choice` into a function that will be evaluated when the
+    * [[EitherTask]] is evaluated at execution time.
```
nh13: not necessarily true, since it is evaluated prior to execution time.  It is evaluated when it has no dependencies.

core/src/main/scala/dagr/core/tasksystem/EitherTask.scala
```diff
@@ -23,24 +23,31 @@
...
+    *
+    * @param left   the left task.
+    * @param right  the right task.
+    * @param choice an expression that returns either Left or Right when evaluated
+    */
+  def apply(left: Task, right: Task, choice: => Choice): EitherTask = new EitherTask(left, right, () => choice)
```
nh13, line 39: cheeky

core/src/main/scala/dagr/core/tasksystem/EitherTask.scala
nh13: if you agree, then fix up all the scaladocs.

core/src/main/scala/dagr/core/tasksystem/EitherTask.scala
```diff
@@ -49,8 +56,9 @@ object EitherTask {
...
   * @param left the left task.
   * @param right the right task
-  * @param flag indicates which task to perform.
+  * @param choice an expression that returns either Left or Right when invoked
   */
-class EitherTask(private val left: Task, private val right: Task, var flag: EitherTaskFlag.Value) extends Task {
-  override def getTasks: Traversable[Task] = List(if (flag == EitherTaskFlag.LEFT) left else right)
+class EitherTask private (private val left: Task, private val right: Task, private val choice: () => EitherTask.Choice) extends Task {
+  /** Decides which task to return based on `choice` at execution time. */
+  override def getTasks: Traversable[Task] = Seq(if (choice() eq EitherTask.Left) left else right)
```
nh13, line 63: `choice()` -> `choice`

nh13:
back to you @tfenne

core/src/main/scala/dagr/core/tasksystem/EitherTask.scala
tfenne: For a task that is not a `UnitTask`, how would you describe the lifecycle?  In my mind, the invocation of `getTasks` _is_ execution time for non-unit tasks like `Either` or `Pipeline`, because that is when their work is done.  

I can change this specific doc to something like `will be evaluated lazily when all dependencies have been met and a choice must be made`, but I'm unclear on why you think the existing wording is wrong.

core/src/main/scala/dagr/core/tasksystem/EitherTask.scala
tfenne, line 63: It has to be `choice()` otherwise it will compare the function reference, not the result of calling the function.

PR #108, opened 2016-02-18
Got rid of the timeout in both stepExecution() and runToCompletion()

tfenne:


core/src/main/scala/dagr/core/execsystem/TaskExecutionRunner.scala
```diff
@@ -270,6 +269,9 @@ private[core] class TaskExecutionRunner extends TaskExecutionRunnerApi with Lazy
       }
   }
 
+  /** Join on all the running processes until they are finished or the timeout reached. */
+  override def joinAll(millis: Long): Unit = this.processes.values.foreach(_.join(millis))
```
nh13: this may join on each for `millis` rather until the timeout is reached.  Update the doc or the desired behavior.

nh13: Also, do you want a default like in the trait?

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -163,6 +163,8 @@ class TaskManager(taskManagerResources: TaskManagerResources = TaskManagerDefaul
   override protected def scriptPathFor(task: Task, taskId: TaskId): Path = pathFor(task, taskId, actualScriptsDirectory, "sh")
   override protected def logPathFor(task: Task, taskId: TaskId): Path = pathFor(task, taskId, actualLogsDirectory, "log")
 
+  def joinOnRunningTasks(millis: Long) = taskExecutionRunner.joinAll(millis)
```
nh13: Either write some doc and consider putting this method into one of the `TaskManager` traits (probably don't), or better yet, make it `private[execsystem]` and indicate it's for testing.

core/src/test/scala/dagr/core/execsystem/TaskExecutionRunnerTest.scala
```diff
@@ -120,6 +120,7 @@ class TaskExecutionRunnerTest extends UnitSpec with OptionValues with BeforeAndA
     // run the task
     taskRunner.runTask(taskInfo = taskInfo)
     taskRunner.runningTaskIds should have size 1 // there should be a running task
+    taskRunner.joinAll()
```
nh13: I still think it makes sense to have some kind of timeout, and fail if not all of them were joined, since I don't want tests that could run forever.

core/src/test/scala/dagr/core/execsystem/TaskExecutionRunnerTest.scala
```diff
@@ -345,6 +347,7 @@ class TaskExecutionRunnerTest extends UnitSpec with OptionValues with BeforeAndA
       logFile = null
     )
     taskRunner.runTask(taskInfo=taskInfo, simulate=false) shouldBe true
+    taskRunner.joinAll()
```
nh13: ditto about running forever.

core/src/test/scala/dagr/core/execsystem/TaskManagerTest.scala
```diff
@@ -536,12 +538,15 @@ class TaskManagerTest extends UnitSpec with PrivateMethodTester with OptionValue
     taskManager.addTask(predecessor)
     taskManager.graphNodeStateFor(predecessor).value should be(GraphNodeState.PREDECESSORS_AND_UNEXPANDED)
     taskManager.stepExecution()
+    taskManager.joinOnRunningTasks(1000)
```
nh13, line 541: Consider mixing in a trait to the task manager created by `getDefaultTaskManager` that overrides the `joinOnRunningTasks` and sets it to `1000` or some reasonable default.

nh13, line 541: we should also make sure that tests will fail if we expected all tasks to complete.

nh13:
@tfenne back to you.  Feel free to do what you want, and then squash and merge (after tests pass).

tfenne:
@nh13 I've addressed most of your feedback - will merge once the tests pass on travis.

PR #111, opened 2016-02-22
Added a fail-fast option and changes to that any tasks that are stopped...

tfenne:
...by dagr during shutdown end in status `STOPPED` instead of `FAILED`.

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
@@ -105,7 +105,8 @@ object TaskManager extends LazyLogging {
           scriptsDirectory: Option[Path] = None,
           logDirectory: Option[Path] = None,
           scheduler: Option[Scheduler] = Some(defaultScheduler),
-          simulate: Boolean = false): BiMap[Task, TaskExecutionInfo] = {
+          simulate: Boolean = false,
+          failFast: Boolean = false): BiMap[Task, TaskExecutionInfo] = {
 
     val taskManager: TaskManager = new TaskManager(
       taskManagerResources = taskManagerResources.getOrElse(defaultTaskManagerResources),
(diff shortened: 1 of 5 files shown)


core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -117,7 +118,7 @@ object TaskManager extends LazyLogging {
     )
 
     taskManager.addTask(task = task)
-    taskManager.runToCompletion()
+    taskManager.runToCompletion(failFast)
```
nh13: name boolean parameters.

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -523,7 +524,7 @@ class TaskManager(taskManagerResources: TaskManagerResources = TaskManagerDefaul
     )
   }
 
-  override def runToCompletion(): BiMap[Task, TaskExecutionInfo] = {
+  override def runToCompletion(failFast: Boolean): BiMap[Task, TaskExecutionInfo] = {
```
nh13, line 527: default this to `false`

core/src/main/scala/dagr/core/execsystem/TaskManagerLike.scala
```diff
@@ -148,7 +148,7 @@ private[execsystem] trait TaskManagerLike {
     *
     * @return a bi-directional map from the set of tasks to their execution information.
     */
-  def runToCompletion(): BiMap[Task, TaskExecutionInfo]
+  def runToCompletion(failFast: Boolean): BiMap[Task, TaskExecutionInfo]
```
nh13, line 151: default here perhaps?

nh13:
back to you @tfenne

PR #112, opened 2016-02-22
Potential fixes for dependent tasks running when dependencies fail.

tfenne:
@nh13 See the new test, which fails if run on master.  I've made two very small tactical edits in `TaskManager`.  Technically I think either one of these is sufficient to fix the problem (though there's one additional test that needs fixing depending on which one you pick as it effects the graph node state of failed tasks).  I don't really like that there's not a graph node state for failures though, it seems wrong to me to leave the graph node in either of `RUNNING` or `COMPLETED`.

Let me know what you think.  Feel free to commit to this branch - though please leave the new test intact and add more if you want more/different tests.

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
@@ -283,7 +283,6 @@ class TaskManager(taskManagerResources: TaskManagerResources = TaskManagerDefaul
 
   /** Gets the set of newly completed tasks and updates their state (to either retry or complete)
     *
-    * @param timeout the length of time in milliseconds to wait for running tasks to complete
     * @return for each completed task, a map from a task identifier to a tuple of the exit code and the status of the `onComplete` method
     */
   private def updateCompletedTasks(): Map[TaskId, (Int, Boolean)] = {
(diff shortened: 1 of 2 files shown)


nh13:
Let me spell out the reasoning for my change (and reversion of one of your changes).
1. Yes, the `TaskStatus` and `GraphNodeState` need a review.  At this very moment, I _like_ not having a failed state for the execution node, since the thing that executes the graph should only care about if it has dependencies or not, and if is running or not.  From our internal conversations, I am likely to change my mind given where we want to go with `TaskManager` and `TaskExecutionRunner`.  Nevertheless, `COMPLETED` means the graph executioner (ha) no longer needs to worry about this node (not true due to `TaskStatus`, hence the need for a review).  But that was the design goal early on.  So I reverted the change `processCompletedTask` to say that the graph node should be set to `COMPLETED` even if the task will remain in a failed state (due to no more retries).
2. The `updatePredecessors` change you made totally fixed a bug.  I modified it slightly to support the `MANUALLY_SUCCEEDED` task status.  Note that this is an example of where my original design mantra from (1) fails, namely `GraphNodeState` and `TaskStatus` are not de-coupled.
3. I had to change your tests, given you told me not to.  Scalatest gives a nice trait called `OptionValues` that allows you to use `option.value` instead of `option.get`, and if the `get` is called on an empty option, the test will fail nicely.  I also made sure that the `GraphNodeState` is set correctly.  Also see my new test.

nh13:
I also wanted to say about the `TaskStatus` and `GraphNodeState` split purpose.  Use `TaskStatus` if you want to know the execution state of the task; use `GraphNodeState` if you want to know the state of the node in the execution graph.  They were meant to be de-coupled, but now are not.

PR #113, opened 2016-02-23
Various new tasks and tweaks to existing tasks.

tfenne:
A bunch of minor tweaks, mostly to tasks.

### DIFF
--- a/commons/src/main/scala/dagr/commons/io/Io.scala
+++ b/commons/src/main/scala/dagr/commons/io/Io.scala
@@ -143,4 +143,11 @@ object Io {
       case parent => findFirstExtentParent(parent)
     }
   }
+
+  /** Writes one or more lines to a file represented by a path. */
+  def writeLines(path: Path, lines: Seq[String]) = {
+    val writer = Io.toWriter(path)
+    lines.foreach(line => writer.append(line).append('\n'))
+    writer.close()
+  }
 }
--- a/src/main/resources/example.conf
+++ b/src/main/resources/example.conf
@@ -25,6 +25,7 @@ bwa.executable   = ${pipeline_dir}/bwa/bwa
 bwa-kit.dir      = ${pipeline_dir}/bwa.kit
 tabix.executable = ${pipeline_dir}/tabix
 verifybamid.executable = ${pipeline_dir}/verifyBamID
+samtools.executable = ${pipeline_dir}/samtools/samtools
 
 # FreeBayes Configuration
 freebayes.executable = ${pipeline_dir}/freebayes         // required for FreeBayes
--- a/tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
+++ b/tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
@@ -1,0 +1,106 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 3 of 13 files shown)


commons/src/main/scala/dagr/commons/io/Io.scala
```diff
@@ -143,4 +143,13 @@ object Io {
       case parent => findFirstExtentParent(parent)
     }
   }
+
+  /**
+  * Writes one or more lines to a file represented by a path.
+  */
```
nh13: single line comment?

commons/src/main/scala/dagr/commons/io/Io.scala
```diff
@@ -143,4 +143,13 @@ object Io {
...
+  def writeLines(path: Path, lines: Seq[String]) = {
+    val writer = Io.toWriter(path)
+    lines.foreach(line => writer.append(line).append('\n'))
+    writer.close()
```
nh13: it's sad there is no analogous way to do this `Source.fromFile("file.txt").getLines`

src/main/resources/example.conf
```diff
@@ -25,6 +25,7 @@ bwa.executable   = ${pipeline_dir}/bwa/bwa
 bwa-kit.dir      = ${pipeline_dir}/bwa.kit
 tabix.executable = ${pipeline_dir}/tabix
 verifybamid.executable = ${pipeline_dir}/verifyBamID
+samtools.executable = ${pipeline_dir}/samtools/samtools
```
nh13, line 28: I think you want `${pipeline_dir}/samtools` (or at least that's what I want)

tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
```diff
@@ -0,0 +1,109 @@
...
+    this(Map[PathToBam,PathToBam] ( in map {b => b -> out.resolve(PathUtil.basename(b) + ".cleaned" + PathUtil.extensionOf(b).getOrElse(""))}:_*),
+         ref=ref,
+         known=known,
+         intervals=intervals)
+    Files.createDirectories(out)
+  }
+
+  /**
+    * Override this and build your pipeline there. Use the 'add' method
+    * to add tasks and make sure to define dependents between them.
```
nh13: dependents -> dependencies?

tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+/** Runs the realigner target creator step of the process. */
+class RealignerTargetCreator(val in : Seq[PathToBam],
+                             val out: FilePath,
+                             val known: Seq[PathToVcf],
+                             ref: PathToFasta,
+                             intervals: Option[PathToIntervals])
+  extends GatkTask(walker="RealignerTargetCreator", ref=ref, intervals=intervals) {
```
nh13, line 83: Perhaps we should have an enum that lists the available walkers instead of giving a string (ex. `"RealignerTargetCreator"`).  Guards against typos?

tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  override protected def addWalkerArgs(buffer: ListBuffer[Any]): Unit = {
+    in.foreach(bam => buffer.append("-I", bam))
+    buffer.append("-o", out)
+    known.foreach(vcf => buffer.append("-known", vcf))
```
nh13, line 88: it is known.

nh13:
do you want to test `writeLines`?

@tfenne minor stuff, feel free to merge when you are happy.

commons/src/main/scala/dagr/commons/io/Io.scala
tfenne: Agreed.  Perhaps if we write more of these we should create a `Destination` class with similar methods?

src/main/resources/example.conf
tfenne, line 28: I have a samtools dir with a samtools executble (and others) in it, but this is just the example anyway.

tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
tfenne: Heh, I think that's copied down from pipeline.  Will fix.

tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
tfenne, line 83: I think this would just add another level of indirection but be no safer - what's to stop you typo'ing the enum value name?

tfenne:
Good point - I added a simple test that roundtrips through `Io.writeLines` and `Source.fromFile.getLines`.

src/main/resources/example.conf
nh13, line 28: this does tend to become canonical, so I disagree.

PR #114, opened 2016-02-24
Made it so that pipelines can prefix and suffix task names easily.

tfenne:
@nh13 Take a look at this.  The change allows pipelines to have a prefix and/or suffix or neither, which is then pre/app-ended to task names when `getTasks` is called.  This means that you can, e.g. have a pipeline for processing a sequencing run with suffix `.AD09F`, and then per-sample pipeline with their own suffixes like `.SpecialSample`, and then your job names will get `.AD09F.SpecialSample` appended to them - which is then used in naming scripts, logs and in various logging/reporting.

A couple of things for discussion:
1. Should pipeline check, and if missing, add `.` before suffixes and after prefixes?  On the upside it makes things much simpler for users, on the downside if you prefer a separator other than `.` you can't.
2. Should `Pipeline` be responsible for making these file-name safe, or is that the responsibility of the code that creates scripts and logs? (I think the latter since task names can also be problematic).

### DIFF
--- a/core/src/main/scala/dagr/core/tasksystem/NoOpInJvmTask.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/NoOpInJvmTask.scala
@@ -26,6 +26,7 @@ package dagr.core.tasksystem
 /**
  * Trivial No-Op task that runs inside the JVM and does nothing.
  */
-class NoOpInJvmTask(name: String) extends SimpleInJvmTask {
+class NoOpInJvmTask(taskName: String) extends SimpleInJvmTask {
+  this.name = taskName
   override def run(): Unit = Unit
 }
--- a/core/src/main/scala/dagr/core/tasksystem/Pipeline.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/Pipeline.scala
@@ -28,7 +28,10 @@ import java.nio.file.Path
 import dagr.commons.util.LazyLogging
 
 /** Simple trait to track tasks within a pipeline */
-abstract class Pipeline(val outputDirectory: Option[Path] = None) extends Task with LazyLogging {
+abstract class Pipeline(val outputDirectory: Option[Path] = None,
+                        private var prefix: Option[String] = None,
+                        private var suffix: Option[String] = None
+                       ) extends Task with LazyLogging {
   private val tasks = new scala.collection.mutable.LinkedHashSet[Task]()
 
   /** A name exposed to sub-classes that can be treated like a Task, to add root tasks. */
@@ -59,6 +62,17 @@ abstract class Pipeline(val outputDirectory: Option[Path] = None) extends Task w
   final override def getTasks: Traversable[_ <: Task] = {
     build()
     tasks.toList.foreach(addChildren)
+
+    // Do the necessary prefixing and suffixing of task names
+    if (prefix.isDefined || suffix.isDefined) {
+      tasks.foreach {
+        case p: Pipeline =>
+          p.prefix = if (p.prefix.isEmpty) prefix else p.prefix.map(prefix.getOrElse("") + _)
+          p.suffix = if (p.suffix.isEmpty) suffix else p.suffix.map(suffix.getOrElse("") + _)
+        case t: Task     => t.name = prefix.getOrElse("") + t.name + suffix.getOrElse("")
+      }
+    }
+
     tasks.toList
   }
 
--- a/core/src/main/scala/dagr/core/tasksystem/Task.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/Task.scala
@@ -153,9 +153,6 @@ object Task {
  *    method will be called until the task no longer wishes to retry or succeeds.
  */
 trait Task extends Dependable {
-  // A string set by the enclosing workflow to distinguish the task
-  private[tasksystem] var contextName : Option[String] = None
-
   /** The unique id given to this task by the execution system, or None if not being tracked. */
   private[core] var _taskInfo : Option[TaskExecutionInfo] = None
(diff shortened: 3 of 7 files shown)


core/src/main/scala/dagr/core/tasksystem/NoOpInJvmTask.scala
```diff
@@ -26,6 +26,7 @@ package dagr.core.tasksystem
 /**
  * Trivial No-Op task that runs inside the JVM and does nothing.
  */
-class NoOpInJvmTask(name: String) extends SimpleInJvmTask {
+class NoOpInJvmTask(taskName: String) extends SimpleInJvmTask {
```
nh13, line 29: why not just:

```
class NoOpInJvmTask(override var name: String) extends SimpleInJvmTask {
```

Seems unnecessary to have `taskName` here.

core/src/main/scala/dagr/core/tasksystem/Task.scala
```diff
@@ -153,9 +153,6 @@ object Task {
  *    method will be called until the task no longer wishes to retry or succeeds.
  */
 trait Task extends Dependable {
-  // A string set by the enclosing workflow to distinguish the task
-  private[tasksystem] var contextName : Option[String] = None
```
nh13, line 157: thanks!

core/src/test/scala/dagr/core/tasksystem/PipelineTest.scala
```diff
@@ -0,0 +1,79 @@
...
+    class TestPipeline extends Pipeline(prefix=Some("before."), suffix=Some(".after")) {
+      override def build(): Unit = root ==> named("hello") ==> named("world")
+    }
+    val ts = new TestPipeline().getTasks.map(_.name).toList.sorted
+    ts shouldBe List("before.hello.after", "before.world.after")
+  }
+
+  it should "stack prefixes and suffixes for nested pipelines" in {
+    class InnerPipeline extends Pipeline(prefix=Some("innerbefore."), suffix=Some(".innerafter")) {
+      override def build(): Unit = root ==> named("hello") ==> named("world")
```
nh13, line 69: I like `named` so much that perhaps we can change `withName` to `named` in `Task`?

nh13:
I am not sure what I like, so let me throw out some alternative suggestions: 
- Instead of a prefix/suffix, how about providing a method instead that transforms the task name.
- If you want to keep the prefix/suffix, how about methods `withPrefix` and `withSuffix`?
- Get rid of modifying the task name altogether, and let the task manager create a "task name" based on the task and its parents?  Thinking of it similarly to the the namespace of a class and the class name.  The "task name" is the "class name" while the "task context name" is the "package name".  I think this was the motivation behind `contextName` in `Task.scala`?  I think having a really long name `let.me.just.append.TaskName.to.all.the.names` makes it hard to figure out what the individual task was without parsing to which pipelines it belongs. 

^ this needs more discussion before proceeding.

RE: 1. Perhaps the prefix and suffix should not include the `.` and there is a way to override the delimiter?

RE: 2: Names should not care about file-name safety, but those who create the files should, so I think it's the responsibility of those creating the logs or scripts (imagine a place where we have no logs or scripts).

core/src/main/scala/dagr/core/tasksystem/Pipeline.scala
```diff
@@ -59,6 +62,17 @@ abstract class Pipeline(val outputDirectory: Option[Path] = None) extends Task w
...
     build()
     tasks.toList.foreach(addChildren)
+
+    // Do the necessary prefixing and suffixing of task names
+    if (prefix.isDefined || suffix.isDefined) {
+      tasks.foreach {
+        case p: Pipeline =>
+          p.prefix = if (p.prefix.isEmpty) prefix else p.prefix.map(prefix.getOrElse("") + _)
+          p.suffix = if (p.suffix.isEmpty) suffix else p.suffix.map(suffix.getOrElse("") + _)
+        case t: Task     => t.name = prefix.getOrElse("") + t.name + suffix.getOrElse("")
```
nh13, line 72: I don't think it is that big of a concern now and more of a caution, but the list of types of tasks that may need to be matched here could grow based on naming of tasks names.

core/src/main/scala/dagr/core/tasksystem/NoOpInJvmTask.scala
nh13, line 29: but I am wrong as usual, it must be `taskName`.

tfenne, line 29: Exactly - tried it, not allowed to override public vars, so sorry :disappointed:

core/src/test/scala/dagr/core/tasksystem/PipelineTest.scala
tfenne, line 69: Sure - although it's probably time we started doing things like deprecating the old method and adding the new one, to make life easier on ourselves and others.

core/src/main/scala/dagr/core/tasksystem/Pipeline.scala
tfenne, line 72: This is an exhaustive match; I guess the question is more should tasks other than Pipelines be allowed to morph the names of tasks that they return, in which case we would need to handle the combinatorial nesting space!

tfenne:
Thanks @nh13.  Honestly, I'm a little torn.  I don't love overwriting `Task.name`.  On the other hand having two separate names in task, `name` and `contextName` also seems error-prone; everywhere that needs a name for a task needs to be careful to use the right one, and it would be easy to forget and just use `name` where `contextName` is more appropriate.

FWIW `contextName` was a leftover from an earlier abortive attempt I made to implement this, and somehow forgot to revert.  That was when I was thinking that it would contain _just_ the context information, and be mangled into the name at runtime - but it lacked the flexibility to use prefix or suffix notation.

PR #116, opened 2016-02-26
Made it so that pipelines can prefix and suffix task names easily.

nh13:
@tfenne here's an example of an alternative implementation of #114.


core/src/main/scala/dagr/core/tasksystem/Pipeline.scala
```diff
@@ -28,7 +28,9 @@ import java.nio.file.Path
 import dagr.commons.util.LazyLogging
 
 /** Simple trait to track tasks within a pipeline */
-abstract class Pipeline(val outputDirectory: Option[Path] = None) extends Task with LazyLogging {
+abstract class Pipeline(val outputDirectory: Option[Path] = None,
```
nh13, line 31: note to self, align the equals as required by the coding style.

PR #117, opened 2016-02-26
allow a pipeline author to set the number of cores for SamToFastq in

nh13:
bwaMemStreamed.

Sometimes I want to set this to zero, sometimes to one.


tfenne:
:+1:

Issue #121, opened 2016-02-28
Pipe should support && and ||


nh13:
This `ShellCommand("echo", "-n", "Hello", "&&", "echo", "World")` will produce the following command: `echo -n Hello '&&' echo World`.  I want it to produce `echo -n Hello && echo World`.  Similarly if I use `||`.   Perhaps we extend `Pipe` to support this, or add support directly into `ShellCommand`?

tfenne:
Perhaps I'm being dumb, but how is that different from:
`ShellCommand("echo", "-n", "Hello") ==> ShellCommand("echo", "World")`

We've built a whole pipeline system based around the idea of conditionally executing dependencies when upstream tasks succeed :open_mouth: .  I suspect you have some more complicated thing you're actually trying to do that might be illustrative...

nh13:
I want this (don't hurt me): `ShellCommand("echo", "-n", "Hello", "&&", "exit", "0")`

tfenne:
Now I'm even more confused - `echo`, at least under bash, always exits with `0`.

nh13:
I want `ShellCommand("/please/ignore/me/I/cant/make/this/public/but/it/exits/with/non-zero", "&&", "exit", "0")`.  Please humor me that you can imagine that I don't care about the success of a shell command (or script).

nh13:
Closing since I can do this:
`ShellCommand("/command/that/returns/exit/code/1", "\n", "exit", 0)`

PR #124, opened 2016-02-29
Added a way to request that individual process task arguments not get escaped.

tfenne:
@nh13 This is an alternative to #121, to let you do nasty things by by-passing the argument escaping on a per-argument basis.  For the record, I think both this and the enhancement described in #121 are a bad idea - both further couple or expose the fact that execution happens through a shell (and even a specific shell with specific semantics).  I think that is undesirable.  

This at least seems more subtle though, and requires a little more effort from users (you'll have to write your own `ProcessTask` subclass in order to access `unescaped`).


nh13:
I agree this is a bad idea.  I think I could just do this:
`ShellCommand("/command/that/returns/exit/code/1", "\n", "exit", 0)`
and my use case should be ok.

Issue #46, opened 2016-02-05
Standardize the output to terminal when Dagr is running


nh13:
Responses so far:
- @tfenne: One really cool (though non-trivial) answer would be an output similar to top, that shows the running and pending processes in a live-updating table, along with some key stats per process.
- @nh13: I think having a top-like one-screen (not multi-tab) interface is reasonable: I think keeping track of where the cursor lives is the challenge, but others have solved that: http://www.tldp.org/HOWTO/Bash-Prompt-HOWTO/x361.html

nh13:
Closing this issue, as we can now work on improving the interactive mode of dagr.

PR #125, opened 2016-02-29
Top-like task status reporting on the console

nh13:
WIP See: #46

### DIFF
--- a/core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
+++ b/core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
@@ -23,7 +23,7 @@
  */
 package dagr.core.cmdline
 
-import java.io.PrintWriter
+import java.io.{PrintStream, ByteArrayOutputStream, PrintWriter}
 import java.nio.file.{Files, Path}
 
 import dagr.commons.util.{LogLevel, LazyLogging, Logger}
@@ -59,7 +59,6 @@ object DagrCoreMain extends Configuration {
       quiet = false
     )
   }
-
 
   /**
     * Main entry point for the class. Parses the args and executes the pipeline. */
@@ -142,7 +141,9 @@ class DagrCoreMain(
   @arg(doc = "Set the memory available to dagr.", common = true)
   val memory: Option[String] = None,
   @arg(doc = "Write an execution report to this file, otherwise write to the stdout", common = true)
-  val report: Option[Path] = None
+  val report: Option[Path] = None,
+  @arg(doc = "Provide an top-like interface for tasks with the give delay in seconds. This suppress info logging.")
+  var interactive: Boolean = false
 ) extends LazyLogging {
 
   // These are not optional, but are only populated during configure()
@@ -168,10 +169,6 @@ class DagrCoreMain(
     try {
       val config = new Configuration { }
 
-      // System and JVM resources
-      val systemCores  = config.optionallyConfigure[Double](Configuration.Keys.SystemCores) orElse this.cores
-      val systemMemory = config.optionallyConfigure[String](Configuration.Keys.SystemMemory) orElse this.memory
-
       // scripts & logs directories
       val scriptsDirectory = pick(this.scriptDir, pipeline.outputDirectory.map(_.resolve("scripts")),
         config.optionallyConfigure(Configuration.Keys.ScriptDirectory))
@@ -185,11 +182,11 @@ class DagrCoreMain(
         if (errors.nonEmpty) throw new ValidationException(errors.toList)
       }
 
+      Logger.level = this.logLevel
+
       // report file
       this.reportPath = pick(report, pipeline.outputDirectory.map(_.resolve("execution_report.txt")), Option(Io.StdOut)).map(_.toAbsolutePath)
       this.reportPath.foreach(p => Io.assertCanWriteFile(p, parentMustExist=false))
-
-      Logger.level = this.logLevel
 
       val resources = SystemResources(cores = cores.map(Cores(_)), totalMemory = memory.map(Memory(_)))
       this.taskManager = Some(new TaskManager(taskManagerResources=resources, scriptsDirectory = scriptsDirectory, logDirectory = logDirectory))
@@ -208,13 +205,31 @@ class DagrCoreMain(
     val taskMan = this.taskManager.getOrElse(throw new IllegalStateException("execute() called before configure()"))
     val report  = this.reportPath.getOrElse(throw new IllegalStateException("execute() called before configure()"))
 
+    val interactiveReporter: Option[TopLikeStatusReporter] = interactive match {
+      case true if !Terminal.supportsAnsi =>
+        logger.warning("ANSI codes are not supported in your terminal.  Interactive mode will not be used.")
+        interactive = false
+        None
+      case true =>
+        val loggerOutputStream = new ByteArrayOutputStream()
+        val loggerPrintStream = new PrintStream(loggerOutputStream)
+        Logger.out = loggerPrintStream
+        Some(new TopLikeStatusReporter(taskMan, Some(loggerOutputStream), print = (s: String) => System.out.print(s)))
+      case false => None
+    }
+    interactiveReporter.foreach(_.start())
+
     taskMan.addTask(pipeline)
     taskMan.runToCompletion(this.failFast)
 
     // Write out the execution report
-    val pw = new PrintWriter(Io.toWriter(report))
-    taskMan.logReport({str: String => pw.write(str + "\n")})
-    pw.close()
+    if (!interactive || Io.StdOut != report) {
+      val pw = new PrintWriter(Io.toWriter(report))
+      taskMan.logReport({ str: String => pw.write(str + "\n") })
+      pw.close()
+    }
+
+    interactiveReporter.foreach(_.shutdown())
 
     // return an exit code based on the number of non-completed tasks
     taskMan.taskToInfoBiMapFor.count { case (task, info) =>
--- a/core/src/main/scala/dagr/core/execsystem/TaskExecutionInfo.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskExecutionInfo.scala
@@ -24,8 +24,9 @@
 package dagr.core.execsystem
 
 import java.nio.file.Path
-import java.time.Instant
+import java.time.{Duration, Instant}
 
+import dagr.commons.util.TimeUtil._
 import dagr.core.DagrDef._
 import dagr.core.tasksystem.Task
 
@@ -61,5 +62,23 @@ class TaskExecutionInfo(var task: Task,
     s"STATUS[$status] ID[$taskId] NAME[${task.name}] SUBMITTED[${submissionDate.getOrElse(na)}]" +
     s" START[${startDate.getOrElse(na)}] END[${endDate.getOrElse(na)}] ATTEMPT[$attemptIndex]" +
     s" SCRIPT[$script] LOGFILE[$logFile]"
+  }
+
+  /** Gets the total execution time and total time since submission, in seconds, or None if the task has not started and ended.  Formats
+    * the durations or return NA otherwise. */
+  def durationSinceStartAndFormat: (String, String) = {
+    durationSinceStart match {
+      case Some((sinceStart, sinceSubmission)) => (formatElapsedTime(sinceStart), formatElapsedTime(sinceSubmission))
+      case _ => ("NA", "NA")
+    }
+  }
+
+  /** Gets the total execution time and total time since submission, in seconds, or None if the task has not started and ended. */
+  private def durationSinceStart: Option[(Long, Long)] = (this.submissionDate, this.startDate, this.endDate) match {
+    case (Some(submission), Some(start), Some(end)) =>
+      val sinceSubmission = Duration.between(submission, end)
+      val sinceStart      = Duration.between(start, end)
+      Some(sinceStart.getSeconds, sinceSubmission.getSeconds)
+    case _ => None
   }
 }
--- a/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
@@ -136,7 +136,7 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
                   scheduler: Scheduler                       = TaskManagerDefaults.defaultScheduler,
                   simulate: Boolean                          = false,
                   sleepMilliseconds: Int                     = 1000
-) extends TaskManagerLike with TaskTracker with TaskStatusReporter with LazyLogging {
+) extends TaskManagerLike with TaskTracker with FinalStatusReporter with LazyLogging {
 
   private val actualScriptsDirectory = scriptsDirectory getOrElse Io.makeTempDir("scripts")
   private val actualLogsDirectory    = logDirectory getOrElse Io.makeTempDir("logs")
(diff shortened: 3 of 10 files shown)


nh13:
It's fun to try it with the dagr script below and `java -jar target/scala-2.11/dagr-0.1.1-SNAPSHOT.jar --top 0.5 --scripts=Testing.dagr -- Testing`

`Testing.dagr`:

``` scala
package dagr.core

import dagr.core.tasksystem.{Pipeline, _}
import dagr.sopt.clp

// NB: output directories must exist
@clp(description = "Testing")
class Testing extends Pipeline {
    val tasks = for (i <- 1 to 60) yield ShellCommand("sleep", if (i < 50) 0 else 5).withName(s"T$i")
    for (i <- 1 until 60) {
      val j = i - 16
      if (j < 0) root ==> tasks(i)
      else tasks(j) ==> tasks(i)
    }
  }
}
```

nh13:
@tfenne needs some more work, but keeping you in the loop on this one.

nh13:
@tfenne bump.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -142,7 +141,9 @@ class DagrCoreMain(
   @arg(doc = "Set the memory available to dagr.", common = true)
   val memory: Option[String] = None,
   @arg(doc = "Write an execution report to this file, otherwise write to the stdout", common = true)
-  val report: Option[Path] = None
+  val report: Option[Path] = None,
+  @arg(doc = "Provide an top-like interface for tasks with the give delay in seconds. This suppress info logging.")
+  val taskTop: Double = -1
```
tfenne: Definitely needs a better name.  I was thinking about this some and I think I would prefer `interactive`, as this could become a more general switch for an _interactive mode_ that not only displays the top-like console, but perhaps allows other things like dropping to a scala shell etc.

tfenne: Also, do you think it's necessary to make the refresh rate configurable vs. making this just a flag arg?

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -189,10 +186,20 @@ class DagrCoreMain(
       this.reportPath = pick(report, pipeline.outputDirectory.map(_.resolve("execution_report.txt")), Option(Io.StdOut)).map(_.toAbsolutePath)
       this.reportPath.foreach(p => Io.assertCanWriteFile(p, parentMustExist=false))
 
-      Logger.level = this.logLevel
+      if (0 < taskTop && !Terminal.supportsAnsi) throw new ValidationException("ANSI codes are not supported in your terminal.  Interactive mode will not function properly.")
```
tfenne: Hrm - perhaps it should just log a warning and fall-back to regular mode?

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -189,10 +186,20 @@ class DagrCoreMain(
...
+
+      // TODO: Perhaps have the top-like interface have a line or two that shows the last info message?
```
tfenne: I like this idea!  Kinda like a news-crawl at the bottom of the screen.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -189,10 +186,20 @@ class DagrCoreMain(
...
+      Logger.level = if (0 < taskTop) LogLevel.Error else this.logLevel
 
       val resources = TaskManagerResources(cores = cores.map(Cores(_)), totalMemory = memory.map(Memory(_)))
-      this.taskManager = Some(new TaskManager(taskManagerResources=resources, scriptsDirectory = scriptsDirectory, logDirectory = logDirectory))
+      this.taskManager = Some(
+        new TaskManager(
+          taskManagerResources=resources,
+          scriptsDirectory = scriptsDirectory,
+          logDirectory = logDirectory,
+          taskTop = taskTop
```
tfenne: This surprised me a bit.  I would have thought it would be good to keep this out of task manager, and instead have the top-like code be in or called from DagrMain.

core/src/main/scala/dagr/core/execsystem/TaskExecutionInfo.scala
```diff
@@ -24,11 +24,33 @@
...
+import java.time.{Duration, Instant}
 
+import dagr.commons.util.TimeUtil._
 import dagr.core.DagrDef._
 import dagr.core.tasksystem.Task
 
+object TaskExecutionInfo {
+
+  /** Gets the total execution time and total time since submission, in seconds, or None if the task has not started and ended. */
+  def getDurationSinceStart(taskInfo: TaskExecutionInfo): Option[(Long, Long)] = (taskInfo.submissionDate, taskInfo.startDate, taskInfo.endDate) match {
```
tfenne: Why not just define these methods on the `TaskExecutionInfo` class?  I don't understand the benefit of having methods in the companion object that just take in and manipulate an instance of the companion class.  (If you do move them to the class, probably drop the `get` prefix :wink: )

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -523,7 +528,29 @@ class TaskManager(taskManagerResources: TaskManagerResources = TaskManagerDefaul
     )
   }
 
+  /** Refreshes the top-like interface */
+  private class TerminalThread(terminal: TopLikeStatusReporter) extends Runnable {
```
tfenne: Best not to name things `Thread` when they implement `Runnable`, as the two are highly related, but not the same thing.

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -523,7 +528,29 @@ class TaskManager(taskManagerResources: TaskManagerResources = TaskManagerDefaul
...
+        }
+      } catch {
+        case e: InterruptedException => // ignore
+      }
+    }
+  }
+
   override def runToCompletion(failFast: Boolean): BiMap[Task, TaskExecutionInfo] = {
+    val terminalThread = if (0 < taskTop) Some(new Thread(new TerminalThread(this))) else None
+    terminalThread.foreach(_.start())
```
tfenne: A couple of thoughts on the handling of the thread.
1. It would good to call `setDaemon(true)` on the terminalThread, so that if something goes wrong and the main thread exits without interrupting the terminalThread, the VM can still exit
2. I think I would be inclined to implement a "tidier" shutdown mechanism for the thread.  I'd suggest adding an AtomicBoolean to TerminalThread, which you can reach in from the outside and flip instead of using `Thread.interrupt()`, and that the thread checks each iteration, and exits if requested.

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
```diff
@@ -0,0 +1,214 @@
...
+import dagr.commons.util.StringUtil._
+import dagr.commons.util.TimeUtil._
+import dagr.core.tasksystem.{InJvmTask, ProcessTask, Task, UnitTask}
+
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+import scala.tools.jline_embedded.{Terminal => JLineTerminal, TerminalFactory}
+
+/** Simple methods for a terminal */
+object Terminal {
```
tfenne, line 42: This feels very weird - because of the way it's implemented you could get a width and a height that are from different points in a resizing operation (e.g. if the user is dragging the window when it's called).  I'd suggest either

``` scala
def dimensions: (Int, Int) = Some(terminal).map(t -> (t.getWidth, t.getHeight)).get
```

Or making `Terminal` a `class` that allocates a new `JLineTerminal` each time it's instantiated, and then having `TopLikeStatusReporter` generate a new `Terminal` each time.

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
```diff
@@ -0,0 +1,214 @@
...
+  def moveDown(numLines: Int): String = s"\033[${numLines}B"
+
+  def moveRight(numColumns: Int): String = s"\033[${numColumns}C"
+
+  def moveLeft(numColumns: Int): String = s"\033[${numColumns}D"
+
+  def clearScreen: String = "\033[2J"
+
+  def eraseToEOL: String = "\033[K"
+
```
tfenne, line 73: Extra newline!

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
```diff
@@ -0,0 +1,214 @@
...
+    lines.map(truncateLine(_, maxColumns)).mkString("\n")
+  }
+
+  /** A row for the report.  Each row is a single task. This should match the column headers returned by [[reportHeader]]. */
+  private def reportRow(taskInfo: TaskExecutionInfo): List[String] = {
+    // get the total execution time, and total time since submission
+    val (executionTime: String, totalTime: String) = TaskExecutionInfo.getDurationSinceStartAndFormat(taskInfo)
+    // The state of execution
+    val graphNodeState = graphNodeStateFor(taskInfo.taskId).map(_.toString).getOrElse("NA")
+    List(
```
tfenne: Do you think it would be worth the added robustness to make the headers/fields an enum (java or scala) and have reportRow return a Map[ReportField, String]?  Would make it much less error prone to re-order, add and remove fields in the future.

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
```diff
@@ -0,0 +1,214 @@
...
+    * @param print the method to use to write task status information, one line at a time.
+    * @param delimiter the delimiter between entries in a row.
+    */
+  def refresh(print: String => Unit, delimiter: String = "  "): Unit = {
+    val taskInfoMap: BiMap[Task, TaskExecutionInfo] = taskToInfoBiMapFor
+    val taskManagerResources = getTaskManagerResources
+    var numLinesLeft = Terminal.height
+
+    // Erase the screen, line-by-line.
+    for (i <- 0 until Terminal.height) {
```
tfenne: For my edification - does the clear screen code I saw above not work?

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
```diff
@@ -0,0 +1,214 @@
...
+    val jvmMemoryPct: Double    = 100.0 * runningTasks.filterKeys(_.isInstanceOf[InJvmTask]).values.map(_.memory.value).sum / taskManagerResources.jvmMemory.value
+
+    // NB: only considers unit tasks
+    val numRunning    = runningTasks.size
+    val numEligible   = readyTasksList.size
+    val numFailed     = unitTasks.count(taskInfo => TaskStatus.isTaskFailed(taskInfo.status))
+    val numDone       = unitTasks.count(taskInfo => TaskStatus.isTaskDone(taskInfo.status))
+    val numIneligible = unitTasks.count(taskInfo => taskInfo.status == TaskStatus.UNKNOWN)
+
+    header.append(s"Cores:         " + f"${taskManagerResources.cores.value}%6.1f" + ", " + f"$systemCoresPct%.1f" + "% Used\n")
```
tfenne: Seems odd on these lines to use a mix of `String.format`, `f""` and string concats with `+`.  Why not just go all the way and use a single `f""` or `String.format`per header line?

tfenne:
Back to you @nh13.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
nh13: I think we can have the refresh rate configurable in the interface itself like top, so I can just make it a flag for now.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
nh13: :+1:

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
nh13: Moved it out, good idea.

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
nh13: I wish scala enums and case classes could be used together.  I want the ordering of scala enums but the ability to have methods on case classes with pattern matching.

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
nh13: I found that it doesn't.

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
nh13: Because I didn't realize `f"..."` gave me a `printf`-like interface. W00t.

nh13:
@tfenne see the last two commits, where I responded to your comments, and added a header row that prints the last log message.

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
```diff
@@ -52,32 +58,64 @@ object Terminal {
...
+  object ReportField extends Enumeration {
+    type ReportField = Value
+    val ID, NAME, STATUS, STATE, CORES, MEMORY,
+    SUBMISSION_DATE, START_DATE, END_DATE,
+    EXECUTION_TIME, TOTAL_TIME, ATTEMPT_INDEX
+    = Value
+  }
 }
 
 /** Mix this trait into a [[dagr.core.execsystem.TaskManager]] to provide the `refresh()` method for a
```
tfenne, line 123: Comment is out of date (no longer a trait).

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
```diff
@@ -52,32 +58,64 @@ object Terminal {
...
+  /** Start the top like status thread */
+  def start(): Unit = {
+    terminalThread.setDaemon(true)
+    terminalThread.start()
+  }
+
+  /** Shutdown the top like status thread */
+  def shutdown(): Unit = {
+    terminalRunnable.terminate()
+    terminalThread.join()
```
tfenne: Maybe add a timeout here so that if for some bizarre reason the thread is wedged it doesn't prevent shutdown?

core/src/main/scala/dagr/core/execsystem/TopLikeStatusReporter.scala
```diff
@@ -94,26 +132,26 @@ trait TopLikeStatusReporter {
...
+      STATE           -> graphNodeState,
+      CORES           -> f"${taskInfo.resources.cores.value}%.2f",
+      MEMORY          -> taskInfo.resources.memory.prettyString,
+      SUBMISSION_DATE -> timestampStringOrNA(taskInfo.submissionDate),
+      START_DATE      -> timestampStringOrNA(taskInfo.startDate),
+      END_DATE        -> timestampStringOrNA(taskInfo.endDate),
+      EXECUTION_TIME  -> executionTime,
+      TOTAL_TIME      -> totalTime,
+      ATTEMPT_INDEX   -> taskInfo.attemptIndex
+    ).map { case (k, v) => (k, v.toString) }
```
tfenne, line 184: So pretty!

tfenne:
@nh13 This looks great.  My only sadness it that test coverage dropped 5%, but this is such a hard thing to try and test.

nh13:
@tfenne coverage went up, merging.

PR #129, opened 2016-03-04
Refactored Io into a trait and object pair to make it easier to extend.

tfenne:
Made sure that all open() type methods route through a pair of methods for easy overriding.
Added a method to read lines from a file that takes a Path as input and uses the standard method to open the file.

### DIFF
--- a/commons/src/main/scala/dagr/commons/io/Io.scala
+++ b/commons/src/main/scala/dagr/commons/io/Io.scala
@@ -26,25 +26,39 @@ package dagr.commons.io
 import java.io._
 import java.nio.file.{Files, Path}
 
+import scala.collection.mutable.ListBuffer
+import scala.io.Source
+
 /**
- * IO Utility class for working with Path objects.
+* Singleton object to provide access to Io utility methods.
+*/
+object Io extends IoUtil
+
+/**
+ * Trait that can be mixed in to make an Io utility object, and can be re-used elsewhere.
  */
-object Io {
+trait IoUtil {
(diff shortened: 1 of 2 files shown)


commons/src/main/scala/dagr/commons/io/Io.scala
```diff
@@ -26,19 +26,27 @@ package dagr.commons.io
...
  */
-object Io {
+trait IoUtil {
   val StdIn   = PathUtil.pathTo("/dev/stdin")
   val StdOut  = PathUtil.pathTo("/dev/stdout")
   val DevNull = PathUtil.pathTo("/dev/null")
 
   /** Creates a new BufferedWriter to write to the supplied path. */
-  def toWriter(path: Path) : BufferedWriter = Files.newBufferedWriter(path)
+  def toWriter(path: Path) : BufferedWriter = new BufferedWriter(new OutputStreamWriter(toOutputStream(path)))
```
nh13: I am curious about the benefit of this change, could you explain?

nh13:
Mostly curious, other than that +1

commons/src/main/scala/dagr/commons/io/Io.scala
tfenne: Yeah - I want all of the `to..()` methods to vector to a single pair of methods that are easy to override and change the behaviour of.  So subclasses/traits can just override `toInputStream()` and `toOutputStream()` and all other methods will pick up those changes.

PR #130, opened 2016-03-05
Add barcode tag options to MarkDuplicates, and actually use the comment

nh13:
and optical duplicates pixel distance in the command line in
MarkDuplicates.


tfenne:
Feel free to merge.  My only reservation is that I feel like `BARCODE_TAG` isn't a great name in `MarkDuplicates` and would suggest maybe aliasing it to `umiTag` or something similar for the classes in dagr.

nh13:
@tfenne take a look at the names now.

PR #131, opened 2016-03-05
Fixed the memory used for Chained and HighAccuracy Downsampling

nh13:
strategies

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/picard/DownsampleSam.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/DownsampleSam.scala
@@ -34,7 +34,7 @@ import scala.collection.mutable.ListBuffer
 
 object DownsamplingStrategy extends Enumeration {
   type DownsamplingStrategy = Value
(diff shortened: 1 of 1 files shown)


tasks/src/main/scala/dagr/tasks/picard/DownsampleSam.scala
```diff
@@ -34,7 +34,7 @@ import scala.collection.mutable.ListBuffer
 
 object DownsamplingStrategy extends Enumeration {
   type DownsamplingStrategy = Value
-  val HighAccuracy, ConstantMemory, Chained = Value
+  val ConstantMemory, Chained, HighAccuracy = Value
```
tfenne, line 37: I guess this is ok, but I would suggest not re-ordering Enumerations in the future unless there's a really strong reason to.

nh13, line 37: It really bothered me that the match was a different order than in the enumeration. I agree generally it's not a good idea.

PR #132, opened 2016-03-05
what's a bupm?

geoffjentry:


geoffjentry:
Do I get a cut of the profits now?

tfenne:
What profits?  :smile:

geoffjentry:
If there aren't any, it makes it easier to hand out large percentage cuts :)

PR #134, opened 2016-03-06
Descriptions and versions were not being displayed in usage messages.

### LINKED ISSUE #133: sopt never displays full sub-command descriptions
@nh13 It would appear that your latest refactor makes it so that `sopt` never prints out the full description for a sub-command from `@clp(description="..."`).  This is true for both the cases where `parseCommandAndSubcommand()` is invoked (like in dagr) and the case when `parseSubCommand()` is invoked (like in `fgbio`).  E.g. neither of the following print out the command/pipeline descriptions:

`java -jar dagr.jar ExamplePipeline --help`
`java -jar fgbio.jar FilterBam --help`

I took a look at fixing this myself and got lost in the logic around where you want things to show up when doing command/subcommand parsing vs. just sub-command parsing, and when `extraUsage` is and isn't provided.  Can you take a look please?

nh13:
@tfenne fixes https://github.com/fulcrumgenomics/dagr/issues/133.  Can you please verify it produces the intended behavior?

The whole testing of usage and error messages for command line programs is a big PITA, and doesn't cover all the cases (obviously).  Perhaps a better set of tests would include a few common usages of `sopt`:
1. A single CLP with options.
2. A CLP with sub-commands.  The CLP has no options but the sub-commands do.  Example: `fgbio`.
3. A CLP with sub-commands, with bot the CLP and the sub-commands having options.  Example: `dagr`.

We save in a test resource file the text we want them to print for usages, then make sure they match fully.  I think the two major test cases are help messages and when the sub-command list is printed for 2 & 3.  

I don't think this PR is blocked by this, but I thought I would start the discussion here, and we can make it into an issue if you agree.

### DIFF
--- a/sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
+++ b/sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
@@ -115,9 +115,6 @@ trait CommandLineParserStrings {
         s"${KRED(UsagePrefix)} ${KBLDRED(commandLineName)} ${KRED(s"[$genericClpNameOnCommandLine] [arguments]")}"
     }
   }
-
-  /** Returns the implementation version string of the given class */
-  protected def version(clazz: Class[_]): String = "Version:" + clazz.getPackage.getImplementationVersion
 }
 
 /** Class for parsing the command line when we also have sub-commands.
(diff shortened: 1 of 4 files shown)


tfenne:
Thanks for fixing this @nh13.  I agree - it might be better to simply enumerate the list of things we want to test, work on the code until we're happy with the outputs, and then save the outputs and test that they are produced.

PR #137, opened 2016-03-09
Nh memory retry

nh13:


nh13:
@tfenne could you take a look?

core/src/main/scala/dagr/core/tasksystem/MemoryRetryTask.scala
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+package dagr.core.tasksystem
+
+import dagr.core.execsystem.{Resource, TaskExecutionInfo, Memory}
+
+/**
+  * A trait to facilitate retrying a task with more memory.
+  */
+trait MemoryRetryTask extends Task with FixedResources {
```
tfenne: Would it work to a) call this just `MemoryRetry` or `MemoryAlteringRetry` and have it self-type to `Task with FixedResources` rather than extending them directly?

core/src/main/scala/dagr/core/tasksystem/MemoryRetryTask.scala
```diff
@@ -0,0 +1,79 @@
...
+  /** Given the current memory, returns the next memory to retry this task with, or None, if the task
+    * should not be retried.
+    */
+  protected def nextMemory(currentMemory: Memory): Option[Memory]
+
+  /** useSystemMemory true if we should use the system memory as the maximum memory, false to use the heap size. */
+  protected def useSystemMemory: Boolean = true
```
tfenne: I think I would suggest getting rid of all this, and perhaps making the trait extend/self-type to `ProcessTask` or even `JarTask`, since it's not clear to me that this could add any value to an InJvmTask - we don't physically limit the memory the task can use from the heap, and we can't dynamically alter the heap for an InJvmTask.

Then `maximumMemory` can just be defaulted to system memory.

core/src/main/scala/dagr/core/tasksystem/MemoryRetryTask.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  /** The maximum memory that this task can use. */
+  protected def maximumMemory: Memory = if (useSystemMemory) Resource.systemMemory else Resource.heapSize
```
tfenne: Heh, and now I'm thinking that `Resource.systemMemory` really needs to be the actual `TaskManagerResources.systemMemory` because otherwise, when we allocate less than all the memory on an instance do dagr, we'll end up pushing tasks to places where they can't execute.

core/src/main/scala/dagr/core/tasksystem/MemoryRetryTask.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  override def retry(taskInfo: TaskExecutionInfo, failedOnComplete: Boolean): Option[Task] = {
```
tfenne: I think this needs to test whether or not the task ran out of memory before doing the doubling.  Perhaps at least define a `def ranOutOfMemory(taskInfo)` that returns `true` if you don't want to have the base trait trawl the log file?

tfenne:
@nh13 Looks nice, but a few thoughts on fine-tuning how things work in the base trait.  Back to you.

core/src/main/scala/dagr/core/tasksystem/MemoryRetryTask.scala
nh13: We can't get rid of `Task` since `retry` relies on returning an `Option[Task]` and `this.requires` returns a `self.type`, so really this trait needs to extend task.  I am not 100% certain, but having a `this: FixedResources =>` self-type annotation wont work with inherited traits, who I think would also need to repeat that statement.  Again, I am not 100% sure on the last one, but I think that's true base on my noodling.

tfenne: One other option is to pull the `retry()` definition out into a separate `Retry` trait, and just extend that?  Then `Task` could either not implement `Retry` by default, or just mix it in and no-op the implementation?

core/src/main/scala/dagr/core/tasksystem/MemoryRetryTask.scala
nh13: I disagree.  An `InJvmTask` could be scheduled with `2G` of memory, try my best to only use `2G` but eventually need more so therefore fail, then require `4G`, and succeed.  Also, we don't physically limit the amount of memory `bwa` uses either, yet we may retry it with parameters to use more memory (I have an old version of `bwa` that can reduce the suffix array compression for faster lookups but more memory).

core/src/main/scala/dagr/core/tasksystem/MemoryRetryTask.scala
nh13: +1

nh13:
@tfenne mind taking another look?

core/src/main/scala/dagr/core/tasksystem/Retry.scala
```diff
@@ -0,0 +1,137 @@
...
+  * A trait to facilitate retrying a task with more memory.
+  */
+trait MemoryRetry extends Retry with FixedResources {
+  /** Given the current memory, returns the next memory to retry this task with, or None, if the task
+    * should not be retried.
+    */
+  protected def nextMemory(currentMemory: Memory): Option[Memory]
+
+  /** True if we should use the system memory as the maximum memory, false to use the heap size. */
+  protected def useSystemMemory: Boolean = true
```
tfenne, line 54: I'm still not sure this is necessary, but I guess it doesn't hurt.

core/src/main/scala/dagr/core/tasksystem/Retry.scala
```diff
@@ -0,0 +1,137 @@
...
+    * @param systemResources the system resources under which this task should run.
+    * @param taskInfo the task execution information for the task to be retried.
+    * @return true if the task is to be retried, false otherwise.
+    */
+  override final def retry(systemResources: SystemResources, taskInfo: TaskExecutionInfo): Boolean = {
+    if (taskInfo.resources.memory != this.resources.memory) throw new IllegalStateException("Scheduled memory does not equal current memory")
+    if (ranOutOfMemory(taskInfo)) {
+      val maximumMemory = if (useSystemMemory) systemResources.systemMemory else systemResources.jvmMemory
+      nextMemory(this.resources.memory)
+        .filter { memory => memory <= maximumMemory }
```
tfenne: I think this would be slightly clearer as just `_ <= maximumMemory`, since my mind wants to see the `=>` and `<=` as symmetrical operators.

core/src/main/scala/dagr/core/tasksystem/Retry.scala
```diff
@@ -0,0 +1,137 @@
...
+/**
+  * A trait for all tasks that wish to linearly increase their memory upon each retry.
+  *
+  *  By default retries until the maximum system memory is reached, and increases by the initial amount of memory required.
+  */
+trait LinearMemoryRetry extends MemoryRetry {
+  private lazy val initialMemory: Memory = this.resources.memory
+
+  /** The maximum amount of memory with which to retry. */
+  def toMemory: Memory = Memory.none
```
tfenne: Doing this will force people to override `toMemory` to make this trait usable right?  Why not either make it abstract so you can't forget to implement it, or initialize it to `Memory.infinite`, which will effectively limit it to the system resources at runtime?

core/src/main/scala/dagr/core/tasksystem/Retry.scala
```diff
@@ -0,0 +1,137 @@
...
+/**
+  * A trait for all tasks that wish to double their memory upon each retry.
+  */
+trait MemoryDoublingRetry extends MemoryRetry {
+  override protected def nextMemory(currentMemory: Memory): Option[Memory] = {
+    Some(currentMemory + currentMemory)
+  }
+}
+
+/** Simple task to retry a given number of times */
```
tfenne: task -> trait

tfenne:
Looks really nice @nh13!  A couple of minor comments which you may do with as you like, then squash and merge.

PR #140, opened 2016-03-11
Moved GenotypeSingleGvcf to GenotypeGvcfs and added apply() methods f…

tfenne:
…or single and multi-sample versions.

Also tweaked Pipeline to create output directory if defined.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/gatk/GenotypeGvcfs.scala
+++ b/tasks/src/main/scala/dagr/tasks/gatk/GenotypeGvcfs.scala
@@ -28,19 +28,31 @@ import DagrDef._
 
 import scala.collection.mutable.ListBuffer
 
-/**
-  * Genotypes a single GVCF.
-  */
-class GenotypeSingleGvcf(ref: PathToFasta,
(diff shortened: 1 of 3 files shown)


tasks/src/main/scala/dagr/tasks/gatk/GenotypeGvcfs.scala
```diff
@@ -28,19 +28,31 @@ import DagrDef._
...
+    new GenotypeGvcfs(ref, Some(intervals), gvcfs, vcf, dbSnpVcf)
+  }
+
+  /** Constructs a GenotypeGvcfs for genotyping a single sample. */
+  def apply(ref: PathToFasta, intervals: PathToIntervals, gvcf: PathToVcf, vcf: PathToVcf, dbSnpVcf: Option[PathToVcf]) : GenotypeGvcfs = {
+    new GenotypeGvcfs(ref, Some(intervals), Seq(gvcf), vcf, dbSnpVcf)
+  }
+}
+
+
```
nh13, line 44: the ocean is vast, and so are newlines ;P

PR #142, opened 2016-03-18
Added mechanism for altering buffer size for reading/writing operations.

tfenne:


commons/src/main/scala/dagr/commons/io/Io.scala
```diff
@@ -42,17 +42,20 @@ trait IoUtil {
   val StdOut  = PathUtil.pathTo("/dev/stdout")
   val DevNull = PathUtil.pathTo("/dev/null")
 
+  /** How large a buffer should be used when buffering operations. */
+  def bufferSize: Int = 32 * 1024
```
nh13, line 46: `Long`?

nh13:
Minor comment, then ready to merge.

commons/src/main/scala/dagr/commons/io/Io.scala
tfenne, line 46: It can't easily be a long - all the underlying APIs take ints, presumably because they are using a byte[] or similar to do buffering, and arrays are limited to length == Int.MAX_VALUE.

PR #144, opened 2016-03-25
Adding support for FgBio tasks and the TrimFastq task.

nh13:


tasks/src/main/scala/dagr/tasks/JarTask.scala
```diff
@@ -27,8 +27,22 @@ import java.nio.file.Path
 
 import dagr.core.execsystem.Memory
 
+import scala.annotation.tailrec
 import scala.collection.mutable.ListBuffer
 
+object JarTask {
+  /** Looks up the first super class that is does not have "\$anon\$" in its name. */
```
tfenne: Extra "is" in the sentence.

tasks/src/main/scala/dagr/tasks/JarTask.scala
```diff
@@ -27,8 +27,22 @@ import java.nio.file.Path
...
+  @tailrec
+  final def findCommandName(clazzOption: Option[Class[_]], errClassName: Option[String] = None): String = {
```
tfenne: It feels a bit weird that the method takes an `Option[Class]` as input - because an outside caller should never invoke it with `None`.

tasks/src/main/scala/dagr/tasks/fgbio/FgBioTask.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import java.nio.file.Path
+
+import dagr.core.config.Configuration
+import dagr.core.tasksystem.{FixedResources, ProcessTask}
+import dagr.tasks.JarTask
+
+import scala.collection.mutable.ListBuffer
+
+
```
tfenne: Extra newline.

tasks/src/main/scala/dagr/tasks/fgbio/FgBioTask.scala
```diff
@@ -0,0 +1,63 @@
...
+  val FgBioJarConfigPath = "fgbio.jar"
+}
+
+/**
+  * Base Task for any task in the FgBio jar.
+  */
+abstract class FgBioTask extends ProcessTask with JarTask with FixedResources with Configuration {
+
+  /** Looks up the first super class that is does not have "\$anon\$" in its name. */
+  def commandName: String = JarTask.findCommandName(Option(getClass), Some("FgBioTask"))
```
tfenne: why make this a `def` that gets invoked every time you need the command name (e.g. in the `args` method), and not just either `[lazy] val commandName`, or even just `this.name = JarTask.....`?

tasks/src/main/scala/dagr/tasks/fgbio/FgBioTask.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  name = commandName
+
+  override def args: Seq[Any] = {
```
tfenne: Should this be final to prevent accidental overriding?

tfenne:
A few minor comments, but otherwise looks good.  Address what you think is relevant, squash and merge.

tasks/src/main/scala/dagr/tasks/JarTask.scala
nh13: I will fix it, but that's the way it was in `PicardTask` previously.

tasks/src/main/scala/dagr/tasks/fgbio/FgBioTask.scala
nh13: Again, that was the way it was in `PicardTask` previously.  Good catch, it can be improved.

tasks/src/main/scala/dagr/tasks/fgbio/FgBioTask.scala
nh13: Should be I think.  I will update `PicardTask` too.

PR #147, opened 2016-04-05
Adding FgBio tool ExtractUmisFromBam.

nh13:
The PR for the tool in fgbio can be found here: https://github.com/fulcrumgenomics/fgbio/pull/14

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/fgbio/ExtractUmisFromBam.scala
+++ b/tasks/src/main/scala/dagr/tasks/fgbio/ExtractUmisFromBam.scala
@@ -1,0 +1,51 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


tasks/src/main/scala/dagr/tasks/fgbio/ExtractUmisFromBam.scala
```diff
@@ -0,0 +1,49 @@
...
+                         val tags2: Seq[String],
+                         val annotateNames: Boolean = true
+                        ) extends FgBioTask {
+
+  protected def addFgBioArgs(buffer: ListBuffer[Any]): Unit = {
+    buffer.append("-i", in)
+    buffer.append("-o", out)
+    buffer.append("-r", rs1)
+    buffer.append("-R", rs2)
+    tags1.foreach(buffer.append("-b", _))
```
tfenne: Probably a matter of taste, but I think I would write:

``` scala
if (tags1.nonEmpty) {
    buffer.append("-b")
    buffer.append(tags1:_*)
}
```

So that it would produce `-b ZA ZB` instead of `-b ZA -b ZB`.

tfenne:
Looks good to me.

PR #148, opened 2016-04-06
Two features in two commits...

tfenne:
One to leverage the new min trimming option in SamToFastq.
One to add support for `--bam_compression` to BAM writing GATK tasks.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
+++ b/tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
@@ -34,40 +34,53 @@ import dagr.tasks.misc.DeleteFiles
 
 import scala.collection.mutable.ListBuffer
 
+object IndelRealignment {
+  // Apply method here since we can't have two constructors in IndelRealignment that take default parameters.
+  def apply(bams: Map[PathToBam, PathToBam],
+            ref: PathToFasta,
+            known: Seq[PathToVcf] = Nil,
+            intervals: Option[PathToIntervals] = None,
+            bamCompression: Option[Int] = None) = {
+    new IndelRealignment(bams=bams, ref=ref, known=known, intervals=intervals, bamCompression=bamCompression)
+  }
+}
+
 @clp(
-  description="Run GATK indel cleaning on one or more BAMs together. Output BAMs are created in the output directory " +
-              "with name = basename(in) + .cleaned. + suffix(in).",
-  hidden=true
+  description = "Run GATK indel cleaning on one or more BAMs together. Output BAMs are created in the output directory " +
+    "with name = basename(in) + .cleaned. + suffix(in).",
+  hidden = true
 )
-class IndelRealignment(val bams: Map[PathToBam,PathToBam],
+class IndelRealignment(val bams: Map[PathToBam, PathToBam],
                        ref: PathToFasta,
                        val known: Seq[PathToVcf],
-                       intervals: Option[PathToIntervals]
+                       intervals: Option[PathToIntervals],
+                       bamCompression: Option[Int]
                       ) extends Pipeline {
 
   /** CLP constructor for testing and manual use, since there's no easy way to provide a map on the CLI. */
-  def this( @arg(flag="i", doc="One or more input BAM files.")       in : Seq[PathToBam],
-            @arg(flag="o", doc="Output directory.")                  out: DirPath,
-            @arg(flag="r", doc="Path to the reference fasta.")       ref: PathToFasta,
-            @arg(flag="k", doc="Zero or more VCFs of known indels.") known: Seq[PathToVcf] = Nil,
-            @arg(flag="l", doc="Optional regions to run over")       intervals: Option[PathToIntervals] = None) = {
-    this(Map[PathToBam,PathToBam] ( in map {b => b -> out.resolve(PathUtil.basename(b) + ".cleaned" + PathUtil.extensionOf(b).getOrElse(""))}:_*),
-         ref=ref,
-         known=known,
-         intervals=intervals)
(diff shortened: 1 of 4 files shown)


tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
```diff
@@ -34,53 +34,66 @@ import dagr.tasks.misc.DeleteFiles
...
+object IndelRealignment {
+  // Apply method here since we can't have two constructors in IndelRealignment that take default parameters.
+  def apply(bams: Map[PathToBam, PathToBam],
+            ref: PathToFasta,
+            known: Seq[PathToVcf] = Nil,
+            intervals: Option[PathToIntervals] = None,
+            bamCompression: Option[Int] = None) = {
+    new IndelRealignment(bams=bams, ref=ref, known=known, intervals=intervals, bamCompression=bamCompression)
+  }
+}
```
nh13, line 46: Why not just have the main constructor have the defaults and get rid of the `apply` method?  I don't see much difference.

tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
```diff
@@ -34,53 +34,66 @@ import dagr.tasks.misc.DeleteFiles
...
-    this(Map[PathToBam,PathToBam] ( in map {b => b -> out.resolve(PathUtil.basename(b) + ".cleaned" + PathUtil.extensionOf(b).getOrElse(""))}:_*),
-         ref=ref,
-         known=known,
-         intervals=intervals)
+  def this(@arg(flag = "i", doc = "One or more input BAM files.") in: Seq[PathToBam],
+           @arg(flag = "o", doc = "Output directory.") out: DirPath,
+           @arg(flag = "r", doc = "Path to the reference fasta.") ref: PathToFasta,
+           @arg(flag = "k", doc = "Zero or more VCFs of known indels.") known: Seq[PathToVcf] = Nil,
+           @arg(flag = "l", doc = "Optional regions to run over") intervals: Option[PathToIntervals] = None) = {
+    this(Map[PathToBam, PathToBam](in map { b => b -> out.resolve(PathUtil.basename(b) + ".cleaned" + PathUtil.extensionOf(b).getOrElse("")) }: _*),
```
nh13: why not just use `toMap` at the end?

tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
```diff
@@ -34,53 +34,66 @@ import dagr.tasks.misc.DeleteFiles
...
+      ref = ref,
+      known = known,
+      intervals = intervals,
+      bamCompression = None)
```
nh13: what, no alignment of the equal signs?

tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
```diff
@@ -34,53 +34,66 @@ import dagr.tasks.misc.DeleteFiles
...
     Files.createDirectories(out)
   }
 
   /** Constructs a mini pipeline that runs RealignerTargetCreator and IndelRealigner on multiple samples together. */
   override def build(): Unit = {
-    val ioMap: FilePath        = Files.createTempFile("indel_cleaning.", ".map")
+    val ioMap: FilePath = Files.createTempFile("indel_cleaning.", ".map")
     val indelCleaningIntervals = Files.createTempFile("indel_cleaning.", ".intervals")
-    val targetCreator = new RealignerTargetCreator(in=bams.keys.toSeq, out=indelCleaningIntervals, known=known, ref=ref, intervals=intervals)
+    val targetCreator = new RealignerTargetCreator(in = bams.keys.toSeq, out = indelCleaningIntervals, known = known, ref = ref, intervals = intervals)
```
nh13, line 78: silly intellij

tasks/src/main/scala/dagr/tasks/picard/SamToFastq.scala
```diff
@@ -46,17 +46,26 @@ class SamToFastq(in: PathToBam,
                  fastq2: Option[PathToFastq] = None,
                  interleave : Boolean = true,
                  clippingAttribute: Option[String] = None,
-                 clippingAction: Option[String] = None) extends PicardTask with Pipe[SamOrBam,Fastq] {
+                 clippingAction: Option[String] = None,
+                 clippingMinReadLength: Int = 25) extends PicardTask with Pipe[SamOrBam,Fastq] {
```
nh13, line 50: It may be overkill, but having `clippingMinReadLength` be an option may make it clearer that attribute/action/minreadlength are set together.

nh13:
Minor stuff.  There are some Intellij spacing changes that if you are so inclined to change do so.  Other than than that, merge when satisfied after the changes.

tasks/src/main/scala/dagr/tasks/gatk/IndelRealigner.scala
tfenne, line 46: Per the comment on the apply method - you can't have two constructors with default parameters, so having this one as an apply method let's me use default parameters.

tasks/src/main/scala/dagr/tasks/picard/SamToFastq.scala
tfenne, line 50: I'd rather leave it like this - mostly because I think that other than for backwards compatibility reasons I would have changed `SamToFastq` in Picard to have this behaviour by default.  I.e. I think it would be too easy to forget and not supply a `clippingMinReadLength`, and I'd rather not have that be a runtime error.

PR #150, opened 2016-04-10
Interval lists should be optional for FreeBayes.

nh13:
This enables whole genome variant calling for FreeBayes.

### DIFF
--- a/pipelines/src/main/scala/dagr/pipelines/GermlineVariantCallingPipeline.scala
+++ b/pipelines/src/main/scala/dagr/pipelines/GermlineVariantCallingPipeline.scala
@@ -28,6 +28,7 @@ import java.nio.file.Path
 import dagr.core.cmdline.Pipelines
 import dagr.sopt._
 import dagr.core.tasksystem.{Linker, Pipeline, ShellCommand}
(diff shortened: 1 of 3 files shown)


pipelines/src/main/scala/dagr/pipelines/GermlineVariantCallingPipeline.scala
```diff
@@ -51,7 +52,8 @@ description =
 class GermlineVariantCallingPipeline
 ( @arg(flag="i", doc="The input BAM file (indexed) from which to call variants.")   val in: PathToBam,
   @arg(flag="o", doc="Output prefix (including directories) for output files.")     val out: Path,
-  @arg(flag="l", doc="Path to interval list of regions to call.")                   val intervals: PathToIntervals,
+  @arg(flag="l", doc="Path to interval list of regions to call (required from GATK).")
```
tfenne: from -> for

tfenne:
Changes look reasonable, but it makes me thing that perhaps we should also modify HaplotypeCaller and friends to take `Option[Intervals]`.  The base task already takes an option, so it would be an easy change.

Also, not sure why the build is failing, but feel free to merge once it's passing.

PR #152, opened 2016-04-10
Adding Picard's RevertSam tool.

nh13:


tasks/src/main/scala/dagr/tasks/picard/RevertSam.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import scala.collection.mutable.ListBuffer
+
+class RevertSam(val in: PathToBam,
+                val out: PathToBam,
+                sortOrder: SortOrder = SortOrder.queryname,
+                restoreOriginalQualities: Boolean = true,
+                removeDuplicateInformation: Boolean = true,
+                removeAlignmentInformation: Boolean = true,
+                attributesToClear: Option[Seq[String]] = None,
```
tfenne: I get why you did this, but `Option[Seq[String]]` feels weird.  How would you feel about making it just `Seq[String]` with a default value of `new picard.sam.RevertSam().ATTRIBUTES_TO_CLEAR.toSeq`?

tasks/src/main/scala/dagr/tasks/picard/RevertSam.scala
```diff
@@ -0,0 +1,59 @@
...
+                santize: Boolean = false)
+  extends PicardTask with PicardMetricsTask {
```
tfenne: I don't think this should include `PicardMetricsTask`.

tasks/src/main/scala/dagr/tasks/picard/RevertSam.scala
```diff
@@ -0,0 +1,59 @@
...
+  requires(Cores(1), Memory("4G"))
+
+  override def metricsExtension: String = MarkDuplicates.MetricsExtension
```
tfenne: And this is probably a copy/paste error?

tasks/src/main/scala/dagr/tasks/picard/RevertSam.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  override protected def addPicardArgs(buffer: ListBuffer[Any]): Unit = {
+    buffer.append("I=" + in)
+    buffer.append("O=" + out)
+    buffer.append("SO=" + sortOrder)
+    buffer.append("RESTORE_ORIGINAL_QUALITIES=" + restoreOriginalQualities)
+    buffer.append("REMOVE_DUPLICATE_INFORMATION=" + removeDuplicateInformation)
+    buffer.append("REMOVE_ALIGNMENT_INFORMATION=" + removeAlignmentInformation)
+    if (attributesToClear.nonEmpty) {
+      buffer.append("ATTRIBUTE_TO_CLEAR=" + null)
```
tfenne: I'm a little leary of passing `null` to `+` here.  How about just `buffer.append("ATTRIBUTE_TO_CLEAR=null")`?

tfenne:
A few minor comments @nh13.  Feel free to squash and merge when you've made the changes you feel are needed.

tasks/src/main/scala/dagr/tasks/picard/RevertSam.scala
nh13: yup

PR #155, opened 2016-04-12
Provide task names for some Germline calling tasks

nh13:


pipelines/src/main/scala/dagr/pipelines/GermlineVariantCallingPipeline.scala
```diff
@@ -71,7 +71,7 @@ class GermlineVariantCallingPipeline
   override def build(): Unit = {
 
     // Ensure the output directory exists
-    val mkdir = ShellCommand("mkdir", "-p", outputDir.toAbsolutePath)
+    val mkdir = ShellCommand("mkdir", "-p", outputDir.toAbsolutePath).withName("Make Output Directory")
```
nh13: I don't think that this task is actually necessary anymore given that `Pipeline` creates the directory.

tfenne: I believe you are correct.

tfenne:
Looks fine to me.  Merge at will.

PR #157, opened 2016-04-18
Allow the user to set the FASTQ quality format when converting to

nh13:
unmapped SAM/BAM.


nh13:
No objections here after a week; I need this; merging.

PR #159, opened 2016-04-26
MarkDuplicates can take a list of input BAMs

nh13:


tasks/src/main/scala/dagr/tasks/picard/MarkDuplicates.scala
```diff
@@ -48,10 +48,12 @@ class MarkDuplicates(override val in: PathToBam,
   extends PicardTask with PicardMetricsTask {
   requires(Cores(1), Memory("6G"))
 
+  def in: PathToBam = inputs.head
```
tfenne: `override`?

tasks/src/main/scala/dagr/tasks/picard/MarkDuplicates.scala
```diff
@@ -48,10 +48,12 @@ class MarkDuplicates(override val in: PathToBam,
...
+
   override def metricsExtension: String = MarkDuplicates.MetricsExtension
 
   override protected def addPicardArgs(buffer: ListBuffer[Any]): Unit = {
-    buffer.append("I=" + in)
+    inputs.tail.foreach(bam => buffer.append("I=" + bam))
```
tfenne: Maybe just comment here that the head of the list is added by the super class to prevent confusing our future selves?

tfenne:
Couple of minor nits - feel free to squash and merge when ready.

PR #166, opened 2016-05-27
Use all available programs but RnaSeqMetrics by default.

nh13:


tfenne:
I hadn't even realized the default was to run all of them.  My preference would also be to exclude CollectGcBiasMetrics by default since that's also not applicable in all cases, but feel free to either make that change or merge as is.

PR #169, opened 2016-06-13
Added a class to consume from an input stream and write to an arbitra…

tfenne:
…ry string consuming function.


tfenne:
@nh13 This is basically identical to the PR in fgbio, with the addition that it has a one-line method in `IoUtil` to make it easier to find in the future.  I'll merge once the checks pass, but feel free to comment if you think anything needs changing.

PR #170, opened 2016-06-21
Cleaned up the dependency DSL, and...

### LINKED ISSUE #107: a ==> b should return a new Dependable(a,b), not b
The behaviour should be:
`a ==> b : Chain(a,b)`
where then
`x ==> Chain(a,b)` adds a dependency from `x` to `a`
`Chain(a,b) ==> x` adds a dependency from `b` to `x`

Also, should probably have accessors for the first and last item in the chain so that it's possible to mimic currently benhaviour with `(a ==> b).last` or similar.

tfenne:
...made it so that a ==> b returns an object representing the chain a ==> b instead of just b.

This fixes #107

### DIFF
--- a/core/src/main/scala/dagr/core/tasksystem/Dependable.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/Dependable.scala
@@ -29,14 +29,53 @@ package dagr.core.tasksystem
   */
 trait Dependable {
   /** Creates a dependency on this dependable, for the provided Task. */
-  def ==> (other: Task) : other.type
+  final def ==> (other: Dependable) : Dependable = {
+    addDependent(other)
+    new DependencyChain(this, other)
+  }
 
-  /** Creates a dependency on this dependable for all the Tasks in the provided MultiTask. */
-  def ==> (others: MultiTask) : MultiTask
+  /** Must be implemented to handle the addition of a dependent. */
+  def addDependent(dependent: Dependable) : Unit
 
   /** Breaks the dependency link between this dependable and the provided Task. */
-  def !=> (other: Task) : other.type
+  def !=> (other: Dependable) : Unit
 
-  /** Breaks the dependency link between this dependable and the provided Tasks. */
-  def !=> (others: MultiTask) : MultiTask
+  /** Returns an object that can be used to manage dependencies that apply to this and the other Dependable. */
+  def :: (other: Dependable) : Dependable = new DependencyGroup(this, other)
+
+  /** Abstract method that must be implemented to return all Tasks associated with this Dependable. */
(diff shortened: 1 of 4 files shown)


core/src/main/scala/dagr/core/tasksystem/Dependable.scala
```diff
@@ -29,14 +29,52 @@ package dagr.core.tasksystem
...
-  def ==> (other: Task) : other.type
+  def ==> (other: Dependable) : Dependable
 
-  /** Creates a dependency on this dependable for all the Tasks in the provided MultiTask. */
-  def ==> (others: MultiTask) : MultiTask
+  /** Breaks the dependency link between this dependable and the provided Task. */
+  def !=> (other: Dependable) : Unit
+
+  /** Returns an object that can be used to manage dependencies that apply to this and the other Dependable. */
+  def :: (other: Dependable) : Dependable = new DependencyGroup(this, other)
```
nh13, line 44: Can you point me to the test that uses `Dependable.::` versus `Multitask.::`?  I am confused about which to use and when, or if I need to understand that?

core/src/main/scala/dagr/core/tasksystem/Task.scala
```diff
@@ -176,33 +176,20 @@ trait Task extends Dependable {
...
+  override def ==>(other: Dependable) : Dependable = {
+    other.toTasks.foreach(t => {
+      t.dependsOnTasks += this
+      this.dependedOnByTasks += t
+    })
 
-  override def ==>(other: MultiTask) : other.type = {
-    other.tasks.foreach(this ==> _)
-    other
+    new DependencyChain(this, other)
```
nh13: I see how you want this to work, perhaps put in a developer note somewhere about what you want the DSL to support given this modification.

nh13:
👍 , though I am starting to not remember all the rules around our DSL, so perhaps we should start a list to formalize later.

core/src/main/scala/dagr/core/tasksystem/Dependable.scala
tfenne, line 44: `Dependable` is a replacement for `Multitask`, which is deleted below.

tfenne:
Thanks.  Part of my goal here was to get all of this into once place (or two if you count all the `Pipe` stuff), instead of having it scattered around.

core/src/main/scala/dagr/core/tasksystem/Task.scala
tfenne: I've modified things so that `==>` is implemented and final in `Dependable` so that the behaviour is guaranteed, and created a new method `addDependent` that is abstract and must be implemented by sub-classes to handle any side-effects.

PR #171, opened 2016-06-21
A different approach to scatter/gather

tfenne:


tfenne:
@nh13 I forked your branch and took a shot at implementing the other scatter/gather API we talked about.  I translated your pipeline to use the new API and after a little testing and bug-fixing it works!  The code behind the API is fairly short, but very confusing and I have a hard time keeping it straight!

Next steps are:
1. Review with you to make sure you still prefer this API
2. Convert all the tests and maybe write some new ones too
3. Fix the fact that, for now, `a ==> Scatter` and `Scatter ==> b`, where `a` and `b` are external/unrelated tasks, doesn't work

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * Provides a set of objects and classes for composing Scatter/Gather tasks and pipelines.
+ */
+object ScatterGather {
+
+  /**
+   * A type of Task that, when run, partitions some kind of input, and produces one or more outputs
+   * of type A that should be parallelized (scattered) over.
+   */
+  trait Scatterer[A] extends Task {
```
nh13: Just like a `Pipe`, I would prefer to have the input type and output type be present.   For example, `Scatterer[PathToVcf, Seq[VariantContext]]` or `Scatterer[PathToInveralList, Seq[IntervalList]]` or `Scatterer[SequenceDictionary, ReferenceSequence]`.  This tells the reader what the input is that is being scattered and possibly transformed.  I mention the latter since it isn't always an input of type `A` to an output of type `Seq[A]`.

  On the other hand, one could argue that it really doesn't matter what the inputs are to a `Scatterer`, just that it produces a sequence of outputs of a fixed type.  Thoughts?

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  /** Provides a factory method to create a Scatter object from a Scatterer. */
+  object Scatter {
+    def apply[A](scatterer: Scatterer[A]): Scatter[A] = {
+      new ScattererWrapper(scatterer)
+    }
+  }
+
+
```
nh13, line 49: I'll forgive you this time for such a blatant error.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+   * of type A that should be parallelized (scattered) over.
+   */
+  trait Scatterer[A] extends Task {
+    def scatters: Seq[A]
+  }
+
+  /** Provides a factory method to create a Scatter object from a Scatterer. */
+  object Scatter {
+    def apply[A](scatterer: Scatterer[A]): Scatter[A] = {
+      new ScattererWrapper(scatterer)
```
nh13: one liner?

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -101,14 +60,29 @@ class ParallelHaplotypeCaller
...
+    val gather = if (filterVcf) {
+      val filter = gt.map(g => new FilterVcf(in=g.vcf, out=replace(g.vcf, ".vcf.gz", ".filtered.vcf.gz")))
+      filter.gather(fs => new GatherVcfs(in=fs.map(_.out), out=output))
+    }
+    else {
+      gt.gather(gts => new GatherVcfs(in=gts.map(_.vcf), out=output))
+    }
+
+    root ==> scatter
+    gather ==> new DeleteFiles(dir)
```
nh13, line 86: It would look better if it was `root ==> scatter ==> gather ==> new DeleteFiles(dir)`.   Also, the files aren't delete as soon as they could be, but this likely would be fine.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+     */
+    def gather[B <: Task](f: Seq[A] => B) : Gather[A,B]
+  }
+
+
+  /**
+   * Represents a gather operation in a scatter/gather pipeline.
+   */
+  sealed trait Gather[A, B <: Task] extends Task {
+    private[ScatterGather] var sources: Option[Seq[A]] = None
```
nh13, line 77: Could we scatter a `Gather`?  For example, we gather to a `Path` and then want to scatter again. 

```
/**
 * Produces a new Scatter from the gathered task.  Takes a function that maps an Task of type B to a Scatter of object of type C. 
 * 
 * The resulting Scatter[C] can be further mapped or gathered (or both!).
 */
def scatter[C](f: B => scatterer[C]): Scatter[C] = throw new NotImplementedException()
```

This is close to using a `Linker` task.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+
+  /** Implementation of a Gather that performs a single All->One gather operation. */
+  private class SingleGather[A,B <: Task](f: Seq[A] => B) extends Gather[A,B] {
+    override def getTasks: Traversable[_ <: Task] = sources match {
+      case None    => throw new IllegalStateException("Gather.getTasks called before sources populated.")
+      case Some(a) => Some(f(a))
+    }
+  }
+
+
```
nh13, line 88: 👎 nl

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+  /**
+  * Implementation of a Scatter that is just a thinly veiled wrapper around the Scatterer
+  * being used to generate the set of scatters/partitions to operate on.
+  */
+  private class ScattererWrapper[A](scatterer: Scatterer[A]) extends Scatter[A] {
```
nh13: why not put this in the `Scatter` object.  I found it hard to find and it is only used there.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+    val subs    = new ListBuffer[SubScatter[B, _ <: Task]]
+    val gathers = new ListBuffer[Gather[B, _ <: Task]]
+
+    /** Defined here for convenience, so it can be invoked on all subs, though in reality only ever called on Secondaries. */
+    def chain(a: A): B = ???
+
+    /** Ensures that gathers are wired in correctly. */
+    protected def connectGathers: Unit = {
+      gathers.foreach(gather => {
+        gather.sources = Some(tasks)
```
nh13: `tasks.toSeq`?

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+ * Provides a set of objects and classes for composing Scatter/Gather tasks and pipelines.
+ */
+object ScatterGather {
+
+  /**
+   * A type of Task that, when run, partitions some kind of input, and produces one or more outputs
+   * of type A that should be parallelized (scattered) over.
+   */
+  trait Scatterer[A] extends Task {
+    def scatters: Seq[A]
```
nh13: Should `scatters` return `Option[Seq[A]]`?  This would allow the `ScatterGather` to detect when `scatters` is called too early, and can then thrown an exception, rather than having each implementation do the guarding.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+      gather
+    }
+  }
+
+  /**
+  * A type of Scatter that is used to represent the first round of Scatter operation after the
+  * scatterer has created the partitions.  It's tasks are generated using the scatter partitions
+  * as input, and is also responsible for chaining together sub-scatters.
+  */
+  private class PrimarySubScatter[A, B <: Task](val scatterer: Scatterer[A], f: A => B) extends Pipeline with SubScatter[A,B] {
```
nh13: I don't think `scatterer` needs to be a `val`, since it is only access inside this class.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  /**
+  * Sub-trait of scatter that exists solely because we need two sub-types that share some common functionality.
+  * Provides the tracking of tasks generated by the Scatter, sub-Scatters that need to be invoked and wired in
+  * when the partitions are available and the set of Gathers to be performed on this stage of the Scatter.
+  */
+  private trait SubScatter[A, B <: Task] extends Scatter[B] {
+    val tasks   = new ListBuffer[B]
+    val subs    = new ListBuffer[SubScatter[B, _ <: Task]]
```
nh13: Could this just be the following?
`val subs    = new ListBuffer[SecondarySubScatter[B, _ <: Task]]`
It may make it clearer that the `PrimarySubScatter` is the one that takes the initial scattered inputs while the `SecondarySubScatter` tasks simply chain along the "scatter" prior to a final gather.

nh13:
@tfenne I did a thorough review (needs tests!) since once I understood what you were doing, I think you were close enough to warrant it.  Perhaps an explanation at the top of the source file showing an example of how all these classes work together at a high-level would be nice too to remind our future selves.  

My only discussion point was around whether or not we should have `Scatterer[A]` or `Scatterer[A,B]`.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne: I sort of agree, that it would be kinda be nice documentation to require `Scatterer`s to declare their input and output types, I decided against it initially because it's not necessary and since this is the one trait that tasks do have to implement, I wanted to keep it as light as possible.  `Scatterer` is really a `Producer[A]`, it's role is to produce partitions over a domain, but we don't really have an opinion about how it does so.  

For example if there are multiple inputs - say a set of intervals and a BAM and a reference - and all are used to compute the output set of intervals, would that be a `Scatterer[(PathToBam, PathToFasta, PathToIntervals), PathToIntervals]`?  Or just a `Scatterer[PathToIntervals,PathToIntervals]`?

Perhaps it's the name that's the problem?  Scatterer does imply taking something and scattering it - perhaps it should be `PartitionProducer` or something else?

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne, line 49: Hah - this is actually on _purpose_!  When I have lots of little classes and traits together in a file I like to have a couple of blank lines between them to make it easier to see where one ends and the next one begins.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne: yup

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
tfenne, line 86: I think this would be fine - though actually it would be:

``` scala
root ==> scatter ==> hc ==> gt ==> gather ==> new DeleteFiles(dir)
```

It's just that `scatter ==> hc ==> gt ==> gather` is auto-wired by the scatter/gather mechanics.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne, line 77: We could do this, but I'm not sure how that's really different than just doing:

``` scala
val gather = scatter.gather(things => new GatherThings(in=things, out=gatheredThing)
val nextScatter = Scatter(new ThingScatterer(in=gatheredThing))
```

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne: Intellij tells me that is a no-op.  I guess tasks.toList would be most defensive - I can do that.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne: Sure - I went back and forth on this.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne: I've made it a `private val`.  FWIW if you access a constructor parameter anywhere other than in the constructor, scala promotes it to a val anyway, so I prefer to declare it that way.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne: Yes!  Thanks for catching that.  At one point with the way I had things this didn't work, but it does now and it does clean things up a bit.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13: I just ask myself what does it scatter?  Perhaps look to other software that implements scatter/gather or map/reduce to find an appropriate name.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13, line 77: `gatheredThing` needs to be known ahead of time in your example, I don't like that.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13: Yeah, just in case it's a silly iterator.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne, line 77: That's only partly true.  More specifically the place where `GatherThings` is going to deposit it's result needs to be known ahead of time.  `gatheredThing` could easily be a simple container object into which `GatherThings` places something (e.g. an `AtomicReference[Thing]`).  But I see your point.  Let's log an issue for this and implement it when we need it?  You're right that's it's totally possible.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne: A little googling didn't turn up much.  Queue isn't helpful (`ScatterFunction` and `GatherFunction`), and there's just not much else out there.  I think probably it does make sense to call it `PartitionProducer` or `PartitionMaker` or something similar.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13, line 77: Lets defer when we have an example, but food for thought.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13: I like the one word name and something short: `Splitter`, `Divider`, `Separator`, `Subdivider`.  We can keep `Scatterer` for now.

As per your input example (`Scatterer[(PathToBam, PathToFasta, PathToIntervals)`), I would say pick what is the primary input data source, and this case likely the `PathToIntervals`.  While not ideal, I don't think we want ``Scatterer[(PathToBam, PathToFasta, PathToIntervals, Int, Int, Int, Int, Float, Float)` where we include all of the parameters to the scatterer.  I think this is more subjective and meant for clarity, rather than a rule.

tfenne: I'm actually tempted to rename it `Partitioner` then, since even though I wrote it I find it a little hard to visually distinguish between `Scatterer` and `Scatter`

tfenne:
@nh13 I've added a  decent test, cleaned up a touch and squashed.  I don't want to add the extra type parameter to `Partitioner` since it's not needed I think it feels like ceremony.  Want to take a final look before I merge?

pipelines/src/main/scala/dagr/pipelines/ParallelHaplotypeCaller.scala
```diff
@@ -0,0 +1,88 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics
```
nh13: LLC?

tasks/src/test/scala/dagr/tasks/ScatterGatherTests.scala
```diff
@@ -0,0 +1,114 @@
...
+import dagr.tasks.ScatterGather.{Partitioner, Scatter}
+import org.scalatest.BeforeAndAfterAll
+
+/**
+  * Tests for the scatter gather task framework.
+  *
+  * Our test involves a "complicated" little scatter/gather that does the following:
+  *   - Scatter a text file by lines
+  *   - Count the words on each line
+  *   - [Gather] Square the counts
```
nh13: perhaps "square each count" to "differentiate it from "sum the counts", where the former is a map and the latter is a gather?

tasks/src/test/scala/dagr/tasks/ScatterGatherTests.scala
```diff
@@ -0,0 +1,114 @@
...
+  *   - Sum the counts
+  *   - [Gather] Sum the squares
```
nh13: this description doesn't jive with what I see below.  How about

```
- [Scatter] Scatter the text file by lines
- [Map] Count the words on each lines
- [Map] Square the count for each line
- [Gather] Sum the counts across all lines.
```

?

nh13: Probably need a test or two that does:
- `Scatter, Map, Gather, Scatter, Map, Gather`
- `Scatter, Map, Gather.  Map on the original Scatter, and another Gather` (is this supported)?

You could log an issue and I'd be happy to take a shot at adding those in (after you merge this PR).

nh13:
minor stuff, read for merge at your discretion, @tfenne back to you.

tasks/src/test/scala/dagr/tasks/ScatterGatherTests.scala
tfenne: On your first comment here - I accidentally a couple of words.  Square and Sum got swapped.  Fixed now.

On your second line comment here - this test actually _is_ that second case.  It scatters, maps to counts and gathers the counts, and then in parallel maps the `counts` to `counts^2` and gathers those.  Or is this not what you meant?

tfenne:
Fixed the minor points.  Let's get this merged and then add more test cases as we need them.

nh13:
+1

nh13:
Thanks!

PR #178, opened 2016-07-19
Tasks for fgbio's GroupReadsByUmi and CallMolecularConsensusReads.

tfenne:


tfenne:
@nh13 just two new task classes

tasks/src/main/scala/dagr/tasks/fgbio/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import scala.collection.mutable.ListBuffer
+
+/** Task to invoke GroupReadsByUmi. */
+class GroupReadsByUmi(val in:  PathToBam,
+                      val out: PathToBam,
+                      val rawTag: Option[String] = None,
+                      val assignTag: String,
+                      val minMapQ: Option[Int] = None,
+                      val strategy: String = "adjacency",
```
nh13: Perhaps here it makes more sense to have an enum, similar to how we have `DownsamplingStrategy` in `DownsampleSam`.

nh13:
One minor comment, fix up (or not), I would though, and then merge.

tfenne:
I've done as you suggested but don't love it.  Should probably, at some point, make it a Java enum in `fgbio` and depend on that here etc.

PR #179, opened 2016-08-04
Expose a few BWA backtrack options.

nh13:
I could change `maxDiffOrMissingProb` to two `Options` and make sure both are not `Some` value, but I went with `Either` since it seemed more "scala".

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/bwa/BwaBacktrack.scala
+++ b/tasks/src/main/scala/dagr/tasks/bwa/BwaBacktrack.scala
@@ -16,12 +16,17 @@ import scala.collection.mutable.ListBuffer
   * BAM file via bwa backtrack/bwa short alignment.
   */
 class BwaBacktrack(val unmappedBam : PathToBam, val mappedBam : PathToBam,
-                   val ref : PathToFasta, val tmpDir: Option[DirPath] = None
+                   val ref : PathToFasta, val tmpDir: Option[DirPath] = None,
+                   val maxDiffOrMissingProb: Option[Either[Int, Float]] = None,
+                   val seedLength: Option[Int] = None,
+                   val pairedEnd: Boolean = true
                 ) extends Pipeline {
   /**
     * Strings together invocations of bwa aln and bwa sampe to generate an output SAM file.
     */
-  override def build(): Unit = {
+  override def build(): Unit = if (pairedEnd) buildPaired() else buildFragment()
+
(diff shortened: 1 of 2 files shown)


tasks/src/main/scala/dagr/tasks/bwa/BwaBacktrack.scala
```diff
@@ -16,12 +16,17 @@ import scala.collection.mutable.ListBuffer
   * BAM file via bwa backtrack/bwa short alignment.
   */
 class BwaBacktrack(val unmappedBam : PathToBam, val mappedBam : PathToBam,
-                   val ref : PathToFasta, val tmpDir: Option[DirPath] = None
+                   val ref : PathToFasta, val tmpDir: Option[DirPath] = None,
+                   val maxDiffOrMissingProb: Option[Either[Int, Float]] = None,
```
tfenne, line 20: It's really annoying that scala's numeric types don't have a common superclass, so you can't just do `Option[Number]`

tasks/src/main/scala/dagr/tasks/bwa/BwaBacktrack.scala
```diff
@@ -16,12 +16,17 @@ import scala.collection.mutable.ListBuffer
...
+                   val seedLength: Option[Int] = None,
+                   val pairedEnd: Boolean = true
                 ) extends Pipeline {
   /**
     * Strings together invocations of bwa aln and bwa sampe to generate an output SAM file.
     */
-  override def build(): Unit = {
+  override def build(): Unit = if (pairedEnd) buildPaired() else buildFragment()
+
+  def buildPaired(): Unit = {
```
tfenne: This should probably be private.

tfenne:
Looks reasonable to me.  Maybe make the build\* methods private?  Merge at will.

tasks/src/main/scala/dagr/tasks/bwa/BwaBacktrack.scala
nh13, line 20: My impression on reading stackoverflow is most people would just say "roll your own", which tells me they don't get it.

tasks/src/main/scala/dagr/tasks/bwa/BwaBacktrack.scala
nh13: Good catch.

PR #184, opened 2016-08-08
Adding clipping attribute option to ExtractUmisFromBam.

nh13:
Adds support for https://github.com/fulcrumgenomics/fgbio/pull/81


tfenne:
Seems fine other than that I mentioned over in fgbio that I think `-x` is an odd choice for the flag.

PR #185, opened 2016-08-13
SplitIntervalsForCallingRegions should not create interval lists that

nh13:
cover too many bases.

Previously, this caused the first interval list in `ParallelHaplotypeCaller` to be two intervals, one of size `N-1` and one of size `N`, and the rest of interval lists having one interval of size `N`.  We want no interval list to cover more than `N` bases in total.  

For more context, we are using the `IntervalList.breakIntervalsAtBandMultiples` which breaks the intervals at a specific base position multiple (ex, 55, 110, 165, 220, ...).  If we ask it to break every `N`bp multiple, and we feed an interval `(1, M)` where `M > N`, we get `(1, N-1), (N, 2*N-1), ...`.  The issue is that the first interval has size `N-1`, and so the previous implementation would see that `N-1<N` and add the second interval to the first interval list (which contained the first interval).  Hopefully the details make sense.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/gatk/SplitIntervalsForCallingRegions.scala
+++ b/tasks/src/main/scala/dagr/tasks/gatk/SplitIntervalsForCallingRegions.scala
@@ -41,7 +41,8 @@ import scala.collection.mutable.ListBuffer
   *
   * If no intervals are given, it will create one or more interval list files per reference sequence.
   *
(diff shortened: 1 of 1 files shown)


tfenne:
Looks good to me.  It might be worth writing some tests for this since a) you found a bug and b) it's more than a simple wrapper around an external tool.

PR #186, opened 2016-08-15
GatherBamFiles and some make some parameters to MarkDuplicates vals.

nh13:
The latter is so that I can use `MarkDuplicates` in a chain of scatter/gather.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/picard/GatherBamFiles.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/GatherBamFiles.scala
@@ -1,0 +1,38 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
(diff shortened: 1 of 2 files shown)


tasks/src/main/scala/dagr/tasks/picard/GatherBamFiles.scala
```diff
@@ -0,0 +1,38 @@
...
+ *
+ */
+
+package dagr.tasks.picard
+
+import dagr.tasks.DagrDef._
+
+import scala.collection.mutable.ListBuffer
+
+class GatherBamFiles(val in: Traversable[PathToBam], val out: PathToBam)
```
tfenne: Should this maybe take a `Seq` instead of a traversable since order definitely matters?

tfenne:
One minor comment, then merge when ready.

Issue #126, opened 2016-02-29
Errors for omitting require arguments should not use prefix 'UserException'


tfenne:
E.g. `UserException: Argument 'run-dir' is required.`. Users of pipelines may not be programmers - it should just say "Error" or "User Error", not print the kind of exception that was generated.

PR #189, opened 2016-08-17
Print better error messages for command line exceptions.

nh13:
Fixes https://github.com/fulcrumgenomics/dagr/issues/126


tfenne:
Seems reasonable @nh13.

PR #190, opened 2016-08-23
Adding a task for VarDictJava and updating htsjdk/picard dependencies.

nh13:


tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -0,0 +1,92 @@
...
+  */
+object VarDictJava extends Configuration {
+  /** Config key for the VarDictJava directory.  This should be the root project directory. */
+  val VarDictJavaDirConfigKey = "vardictjava.dir"
+
+  /** Path to the VarDictJava directory. */
+  lazy val VarDictJavaDir = configure[Path](VarDictJavaDirConfigKey)
+
+  /** Path to the VarDictJava install directory with the VarDictJava startup script. */
+  lazy val VarDictJava = VarDictJavaDir.resolve("build/install/VarDict/bin/VarDict")
```
tfenne: I think this is the layout you get if you build from source.  But if you download a release package, you just get  `lib` and `bin` directories under the main directory.  That's probably more common.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  /** Path to the VarDictJava bin directory with a multitude of perl scripts. */
+  lazy val VarDictJavaBinDir = VarDictJavaDir.resolve("VarDict")
```
tfenne: I found the naming confusing ... since it's nested in an object called VarDictJava how about just `BinDir`?  Also same comment here about directory structure.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/**
+  * Tumor-only single-sample variant calling with VarDictJava: https://github.com/AstraZeneca-NGS/VarDictJava
+  */
+class VarDictJava(tumorBam: PathToBam,
+                  bed: FilePath,
+                  ref: PathToFasta,
+                  af: Double = 0.01,
```
tfenne: This should probably be `minimumAf`

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -0,0 +1,92 @@
...
+                  out: PathToVcf = Io.StdOut
+                 ) extends Task with Configuration {
```
tfenne: FWIW when I use VarDict elsewhere I also sometimes use:
`-F 0x700` to exclude non-PF reads on top of the default duplicates and secondaries
`-q 30` to limit to Q30+ input bases
`-th 4` to multi-thread

I think it would be good to expose those if possible, although the first one maybe just hardcode or parameterize as `includeNonPf: Boolean = false`?

tfenne: Why not just have it extend pipeline?

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -0,0 +1,92 @@
...
+
+    vardict.requires(Cores(1), Memory("8G"))
+    bias.requires(Cores(0), Memory("32m"))
+    toVcf.requires(Cores(0), Memory("32m"))
+
+    Seq((vardict | bias | toVcf | removeRefEqAltRows > out).withName(this.name))
+  }
+
+  /** Pulls a sample name out of a BAM file. */
+  private def extractSampleName(bam: PathToBam) : String = {
```
tfenne: FYI, the tumor name you supply to vardict can be anything - it's just what it writes to the output file.  So it's fine to extract it like this, or you could just use `tumor` or let the user provide it.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: For posterity, the releases don't contain the perl scripts, so we want to use the source.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: Fixed.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: Fixed.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: Add the options, I want to be able have it return `PipeOut[Vcf]` instead. Will update.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: Adding as an option.

nh13:
@tfenne take a look.  Unfortunately the directory structure is what it is.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -44,44 +46,55 @@ import htsjdk.samtools.SamReaderFactory
...
+    val toVcf              = new ShellCommand(vcfScript.toString, "-N", tn, "-E", "-f", minimumAf.toString) with Pipe[Any,Vcf]
     val removeRefEqAltRows = new ShellCommand("awk", "{if ($1 ~ /^#/) print; else if ($4 != $5) print}") with Pipe[Vcf,Vcf]
 
-    vardict.requires(Cores(1), Memory("8G"))
+    vardict.requires(Cores(threads), Memory("8G"))
     bias.requires(Cores(0), Memory("32m"))
     toVcf.requires(Cores(0), Memory("32m"))
 
-    Seq((vardict | bias | toVcf | removeRefEqAltRows > out).withName(this.name))
+    (vardict | bias | toVcf | removeRefEqAltRows).withName(this.getClass.getSimpleName).asInstanceOf[PipeOut[Vcf]]
```
nh13: It would be really nice if I could avoid the `asInstanceOf` call here.

core/src/main/scala/dagr/core/tasksystem/Pipe.scala
```diff
@@ -118,6 +118,12 @@ object Pipes {
     this.requires(cores, memory)
   }
 
+  /** A type alias fdagr.core.tasksystemor sink tasks that can receive data via a pipe, but cannot pipe onwards. */
+  type PipeIn[In] = dagr.core.tasksystem.Pipe[In,Void]
+
+  /** A type alias for generate tasks that can pipe out, but cannot receive data from a pipe. */
+  type PipeOut[Out] = dagr.core.tasksystem.Pipe[Void,Out]
+
```
nh13: @tfenne check this out.  Since I cannot access `Void` outside of here (that's good), but my method would return `Pipe[Void,Out]`, I need this alias.  I can't use `PipeOut[Out]`.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/** Runs VarDictJava to produce its intermediate tabular format. */
+private class VarDictJava(tumorBam: PathToBam,
+                          bed: FilePath,
+                          ref: PathToFasta,
+                          tumorName: String,
+                          minimumAf: Double = 0.01,
+                          includeNonPf: Boolean = false,
+                          minimumQuality: Int = 0,
```
tfenne: VarDict's default for this is Q25, so I think we should probably either mirror that here, or preferably make this an Option[Int] with None as the default.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -0,0 +1,148 @@
...
+                          maximumMismatches: Int = 8,
```
tfenne: Similarly I'd be temped to make this Option[Int] rather than reproduce the VarDict default here.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -0,0 +1,148 @@
...
+  override def pickResources(resources: ResourceSet): Option[ResourceSet] = {
+    resources.subset(minCores = Cores(minThreads), maxCores = Cores(maxThreads), memory = memory)
+  }
+
+  override def args: Seq[Any] = {
+    val vdScript = VarDictJava.VarDictJava
+    val buffer   = new ListBuffer[Any]()
+
+    buffer.appendAll(List(vdScript, "-G", ref, "-N", tumorName, "-b", tumorBam))
+    buffer.appendAll(List("-z", "-c", "1", "-S", "2", "-E", "3", "-g", "4"))
```
tfenne: Isn't `-z` supposed to take either `0` or `1` as an option value?

Also, can I suggest that you break these out onto multiple lines and add a short comment on each line describing what the option does?  Will save going back and forth to the usage all the time to remember what they all do.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -0,0 +1,148 @@
...
+  * The output Tumor name will be inferred from the the first read group in the input BAM if not provided.
+  */
+class VarDictJavaEndToEnd(tumorBam: PathToBam,
+                          bed: FilePath,
+                          ref: PathToFasta,
+                          out: PathToVcf = Io.StdOut,
+                          tumorName: Option[String] = None,
+                          minimumAf: Double = 0.01,
+                          includeNonPf: Boolean = false,
+                          minimumQuality: Int = 0,
```
tfenne: If these were `Options` you wouldn't have to repeat the hard-coded default here ;)

tfenne:
A few minor things @nh13 .  Merge when you're happy.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: Done.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: Done.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: Done.

PR #192, opened 2016-08-23
Bwa mem should run on frags.

nh13:


tfenne:
I don't see why this is necessary.  `-p` is for "smart pairing" - i.e. if you hand it a single fastq it should auto-detect whether adjacent reads are paired or not.  I'm 99% sure I've used this on unpaired data just fine.  I would really like to avoid having to add an option to the task to tell it whether data is paired or not.

nh13:
@tfenne I'll check again, but unpaired data wasn't working for me.  If you believe it should work, then probably it is PEBKAC.

PR #193, opened 2016-08-25
Delete multiple BAMs at the same time!

nh13:


tasks/src/main/scala/dagr/tasks/picard/DeleteBam.scala
```diff
@@ -34,18 +34,20 @@ import scala.collection.mutable.ListBuffer
   * Class to remove a BAM file and also the BAI file if it exists.  Succeeds if the files can be deleted,
   * *or* if no files were present to be deleted.
   */
-class DeleteBam(val bam: Path) extends SimpleInJvmTask {
-  name = "DeleteBam." + PathUtil.basename(bam)
+class DeleteBam(val bam: Path*) extends SimpleInJvmTask {
+  name = "DeleteBam" + (if (bam.length == 1) "." + PathUtil.basename(bam.head) else "s." + bam.length)
```
tfenne: This will create names like `DeleteBams.3` right?  Which is a bit confusing since the log and script files get task ID suffixes as well.  Maybe `Delete3Bams`?

tasks/src/main/scala/dagr/tasks/picard/DeleteBam.scala
```diff
@@ -34,18 +34,20 @@ import scala.collection.mutable.ListBuffer
...
-    val paths = ListBuffer(bam)
-    PathUtil.extensionOf(bam) match {
-      case Some(".bam") =>
-        paths += PathUtil.pathTo(bam.toString + ".bai")
-        paths += PathUtil.replaceExtension(bam, ".bai")
-      case Some(".cram") =>
-        paths += PathUtil.pathTo(bam.toString + ".crai")
-      case _ =>
+    val paths = ListBuffer[Path]()
+    bam.foreach { b =>
```
tfenne, line 42: Um, does this actually delete the BAM?  It used to get hit because it was in the constructor to the `ListBuffer`, but I don't see it now!  Perhaps you want `val paths = ListBuffer[Path](bam:_*)`?

nh13: 👍

nh13:
@tfenne one last peek since I missed a biggie last time?

tfenne:
Looks good.  👍

PR #196, opened 2016-08-25
Expose some methods for custom task managers.

nh13:


tfenne:
Seems reasonable.  I don't love making `taskExecutionRunner` as I feel like that's one part of the API we'll want to change in future, but so long as this doesn't become a reason not to make changes later, I'm on board.

nh13:
Will expose just the method I need on `taskExecutionRunner` as a `protected` method.

Issue #191, opened 2016-08-23
Support clearing an argument of type Option[String] that has a default value


nh13:
Consider an option:

```
@arg(flag="o", doc="optional option") val arg: Option[String] = Some("Default")
```

I want to set it to `None`, how do I do that?

tfenne:
Agreed, this would be very useful.

PR #197, opened 2016-08-26
Add the ability to clear an option or collection with a special token.

nh13:
I contemplated adding the ability to append to the default values of a collection, but I don't have a use case at the moment, so I am deferring.  Implements #191.

### DIFF
--- a/commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
+++ b/commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
@@ -49,6 +49,8 @@ object ReflectionUtil {
   /** A runtime mirror for when it's necessary. */
   private val mirror: ru.Mirror = scala.reflect.runtime.currentMirror
 
+  val SpecialEmptyOrNoneToken = ":none:"
+
   /**
     * Creates a new instance of various java collections.  Will inspect the type given to see
     * if there is a no-arg constructor and will use that if possible.  If there is no constructor
@@ -290,7 +292,12 @@ object ReflectionUtil {
     */
   private[reflect] def typedValueFromString(resultType: Class[_], unitType: Class[_], value: String*): Try[Any] = Try {
     if (ReflectionUtil.isCollectionClass(resultType)) {
-      val typedValues = value.map(v => buildUnitFromString(unitType, v).get).asInstanceOf[Seq[java.lang.Object]]
+      val typedValues = if (value.length == 1 && value.head.toLowerCase == SpecialEmptyOrNoneToken) {
+        Seq.empty
+      }
+      else {
+        value.map(v => buildUnitFromString(unitType, v).get).asInstanceOf[Seq[java.lang.Object]]
+      }
 
       // Condition for the collection type
       if (ReflectionUtil.isJavaCollectionClass(resultType)) {
@@ -313,6 +320,9 @@ object ReflectionUtil {
     }
     else if (value.size != 1) {
       throw new ReflectionException(s"Expecting a single argument to convert to ${resultType.getSimpleName} but got ${value.size}")
+    }
+    else if (resultType == classOf[Option[_]] && value.head.toLowerCase == SpecialEmptyOrNoneToken) {
+      None
     }
     else {
       buildUnitOrOptionFromString(resultType, unitType, value.head).get
--- a/sopt/src/main/scala/dagr/sopt/README.md
+++ b/sopt/src/main/scala/dagr/sopt/README.md
@@ -82,3 +82,8 @@ or multiple values one after each other:
 - A single value option must have a value after its option name.
 - A multiple value option must have one or more values after its option name.
 
+## Special Types
+
+Arguments of type `Option` can be cleared using the special `:None:` (case-sensitive) token.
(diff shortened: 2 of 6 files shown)


nh13:
@tfenne since `---` didn't work, I am using `//` as the special token for an empty collection or `None` for an `Option`.

sopt/src/main/scala/dagr/sopt/README.md
```diff
@@ -82,3 +82,8 @@ or multiple values one after each other:
 - A single value option must have a value after its option name.
 - A multiple value option must have one or more values after its option name.
 
+## Special Types
+
+Arguments of type `Option` can be cleared using the special `<None>` token.
```
tfenne: I think, looking at the code, this is incorrect.  I think it only responds to `//` for both `Option` and `Collection` arguments.

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -290,7 +292,11 @@ object ReflectionUtil {
     */
   private[reflect] def typedValueFromString(resultType: Class[_], unitType: Class[_], value: String*): Try[Any] = Try {
     if (ReflectionUtil.isCollectionClass(resultType)) {
-      val typedValues = value.map(v => buildUnitFromString(unitType, v).get).asInstanceOf[Seq[java.lang.Object]]
+      val typedValues = value.indexOf(SpecialEmptyOrNoneToken) match {
+        case -1   => value.map(v => buildUnitFromString(unitType, v).get).asInstanceOf[Seq[java.lang.Object]]
+        case idx => value.drop(idx+1).map(v => buildUnitFromString(unitType, v).get).asInstanceOf[Seq[java.lang.Object]]
+      }
+      //val typedValues = value.map(v => buildUnitFromString(unitType, v).get).asInstanceOf[Seq[java.lang.Object]]
```
tfenne: Kill commented out line.

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -290,7 +292,11 @@ object ReflectionUtil {
     */
   private[reflect] def typedValueFromString(resultType: Class[_], unitType: Class[_], value: String*): Try[Any] = Try {
     if (ReflectionUtil.isCollectionClass(resultType)) {
-      val typedValues = value.map(v => buildUnitFromString(unitType, v).get).asInstanceOf[Seq[java.lang.Object]]
+      val typedValues = value.indexOf(SpecialEmptyOrNoneToken) match {
```
tfenne: So .. given that for both `Option` and `Collection` types providing one or more values replaces the default values, I think it would be sufficient (and clearer/simpler/safer) here to check and see if the input `value` has a single value in it which matches the `SpecialEmptyOrNoneToken` or not.  If it does, return `Seq.empty()`, else convert the entire array.

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -324,9 +330,14 @@ object ReflectionUtil {
     * it will either return `None` if `s == null`, or `Some` object wrapping the an object of the wrapped type.
     */
   private def buildUnitOrOptionFromString(resultType: Class[_], unitType: Class[_], value: String): Try[Any] = Try {
-    val unit = buildUnitFromString(unitType, value).get
-    if (resultType == classOf[Option[_]]) Option(unit)
-    else unit
+    if (resultType == classOf[Option[_]] && value == SpecialEmptyOrNoneToken) {
```
tfenne: I don't think this change should be necessary.  `buildUnitOrOptionFromString` is only called from `typedValueFromString`, and that will never pass in the `SpecialEmptyOrNoneToken`.

tfenne (commented):
Looks good mostly, though a few suggestions.

I still don't love using `//` as the token.  Some alternative suggestions:
- `:None:` / `:none:`
- `_None_` / `_none_`
- `None` / `none`

I like `None` surrounded by some token character as you can see!

sopt/src/main/scala/dagr/sopt/README.md
nh13: Fixed.

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
nh13: This wont work since the default values are passed into this method along with `:None:` and the values that should overwrite the default values.

nh13: This basically comes from the desire to _add_ the defaults to _overwriting_ the defaults.  It may seem funny that it trickles down to here, but so be it.

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
nh13: It will if it is an option.  I'll bubble this logic up.

nh13:
@tfenne back to you.  It might be worth a discussion around where defaults are stored and how new values are appended (and now overwritten) as there is logic in `ClpArgument` and here.

PR #198, opened 2016-08-26
Expose joinOnRunningTasks in TaskManager to enable testing of custom …

nh13:
…TaskManagers.

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2015 Fulcrum Genomics LLC
+ * Copyright (c) 2015-2016 Fulcrum Genomics LLC
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -141,7 +141,7 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
   private val actualScriptsDirectory = scriptsDirectory getOrElse Io.makeTempDir("scripts")
   protected val actualLogsDirectory  = logDirectory getOrElse Io.makeTempDir("logs")
 
-  protected val taskExecutionRunner: TaskExecutionRunnerApi = new TaskExecutionRunner()
+  private val taskExecutionRunner: TaskExecutionRunnerApi = new TaskExecutionRunner()
```
nh13, line 144: I use the `terminateTask` instead (from #196) and forgot to make it private again.

core/src/test/scala/dagr/core/execsystem/TaskManagerTest.scala
```diff
@@ -44,18 +44,22 @@ object TaskManagerTest {
   }
 }
 
-class TaskManagerTest extends UnitSpec with PrivateMethodTester with OptionValues with LazyLogging with BeforeAndAfterAll {
+trait TestTaskManager extends TaskManager {
+  override def joinOnRunningTasks(millis: Long) = super.joinOnRunningTasks(millis)
```
nh13, line 48: Silly stuff since I can’t override a method that has `protected[scope]` outside of that package scope.

tfenne:
Looks good to me.

Issue #158, opened 2016-04-22
Command line options should indicate that they are collections


nh13:
Take for example `DnaResequencingFromFastqPipeline`.  The fastq argument:

```
--fastq1=PathToFastq          Input fastq file (optionally gzipped) for read 1.
```

doesn't look like you can pass in more than one fastq, but you can.  Perhaps there should be some symbol for how many can be passed in?

Zero or more:

```
--fastq1=PathToFastq*          Input fastq file (optionally gzipped) for read 1.
```

One or more:

```
--fastq1=PathToFastq+         Input fastq file (optionally gzipped) for read 1.
```

M to N:

```
--fastq1=PathToFastq{N,M}          Input fastq file (optionally gzipped) for read 1.
```

tfenne:
Implemented by @nh13

PR #199, opened 2016-08-27
The # of elements for a collection are shown in the usage.

nh13:
Implements https://github.com/fulcrumgenomics/dagr/issues/158

Likely https://github.com/fulcrumgenomics/dagr/pull/188 should be merged before this one.


tfenne:
Looks really nice.  The only suggestion I have is that `collectionDescription` and `makeCollectionDescription` make it sound like it would be the full description of the option - perhaps call them something like `collectionArity` or `collectionArityString`?

Merge when you're happy.

PR #209, opened 2016-09-02
Define an argument as not having a value if it is an Option that is N…

nh13:
…one.


tfenne:
This looks fine to me.

nh13:
Verified that `Option`s are still optional with this dagr script:

```
package dagr.pipelines

import dagr.core.cmdline.Pipelines
import dagr.core.tasksystem.{Pipeline, SimpleInJvmTask}
import dagr.sopt._

/**
  * Hello World!
  */
@clp(description="Hello World Pipeline.", group = classOf[Pipelines])
class HelloWorldPipeline
( @arg(flag="i", doc="Input string.")        val input: Option[String]
) extends Pipeline {

    root ==> new SimpleInJvmTask {
      override def run(): Unit = {
        System.err.println(input.getOrElse("Input was empty"))
      }
    }
  }
}
```

PR #210, opened 2016-09-03
Adding a function cleanly to perform scala try-with-resources.

nh13:
@tfenne I have found this useful elsewhere and I hoping you will find it useful.  It is basically a scala version of Java's try-with-resources.

### DIFF
--- a/commons/src/main/scala/dagr/commons/CommonsDef.scala
+++ b/commons/src/main/scala/dagr/commons/CommonsDef.scala
@@ -24,6 +24,12 @@
 
 package dagr.commons
 
+import java.io.Closeable
+
+import dagr.commons.util.Logger
+
+import scala.util.{Failure, Success, Try}
+
 /**
   * Object that is designed to be imported with `import CommonsDef._` in any/all classes
   * much like the way that scala.PreDef is imported in all files automatically.
@@ -50,6 +56,7 @@ class CommonsDef {
     *   `val x = foo; foo +=1; return x`
     * to be replaced with
     *   `yieldAndThen(foo) {foo +=1}`
+    *
     * @param it the value to be returned/yielded
     * @param block a block of code to be evaluated
     * @tparam A the type of thing to be returned (usually inferred)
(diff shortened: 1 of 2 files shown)


tfenne:
@nh13 I like the idea of having this, but there are a few things that I would like to discuss design-wise before adding this to CommonsDef:
1. Requiring a cleanup method feels cumbersome to me if the majority of use cases will involve objects that have a close() method.  Maybe we have two versions - one which takes a cleanup method and one while only takes types that either implement Closeable or have a close() method on them, and the uses CloserUtil or similar to do the cleanup?
2. Why `cleanly` as opposed to `tryWith` or something like that?  If the pattern is called "try-with-resources" then I think I (and others) would be more likely to find and use the method if the name were similar.
3. How attached are you to the Try vs. try? I still can't decide if I like the Try/Success/Failure model better than just doing try/catch.

nh13:
@tfenne 1&2 I have no disagreement.  

For 3, you can safely ignore the return and it will "do the right thing", namely call the `cleanup` method in the presence of the exception. htsjdk's `CloserUtil` and dagr's `safelyClose` both have `void/Unit` return types, where as this method returns a `Try`.  Imagine if the former functions returned a boolean or `Option`.  Or do you mean the `try` within the `cleanly` method itself?

nh13:
@tfenne back to you.

commons/src/main/scala/dagr/commons/CommonsDef.scala
```diff
@@ -72,6 +79,63 @@ class CommonsDef {
...
+    * @param doWork the work to perform on the resource, returning a result of type [[B]].
+    * @tparam A the resource type.
+    * @tparam B the result type of the work performed.
+    * @return [[Success]] if the work was performed successfully, [[Failure]] otherwise.
+    */
+  def cleanly[A, B](resource: A, logger: Option[Logger] = None)(cleanup: A => Unit)(doWork: A => B): Try[B] = {
+    try {
+      Success(doWork(resource))
+    } catch {
+      case e: Exception => Failure(e)
```
tfenne: Should this be `Throwable`?

commons/src/main/scala/dagr/commons/CommonsDef.scala
```diff
@@ -72,6 +79,63 @@ class CommonsDef {
...
+    }
+    finally {
+      try {
+        if (resource != null) cleanup(resource)
+      } catch {
+        case e: Exception =>
+          logger match {
```
tfenne: This could be just `logger.foreach(_.exception(e))`

tfenne (approved):
Looks good.  My only real feedback is:
- I would prefer both methods to be called `tryWith` but if you really want `cleanly`, ok.
- If you didn't want to pass in the Logger (e.g. just use one internal to the class) then the first version could take `A*` instead of `A` for the resource, and support multiple `Closeable`s in one use.

nh13:
I changed the name to `tryWith` but I would like the logger optional so that I can log the exception.  I'll merge once the tests pass.

PR #213, opened 2016-09-07
Resource error messages

nh13:
First commit error message:

```
[2016/09/07 19:14:00 | TaskManager | Error] There are 1 tasks ready to be scheduled but not enough system resources available.
[2016/09/07 19:14:00 | TaskManager | Error] Task with name '<redacted>' requires 3.2 cores and 4g memory (task schedulable type: Schedulable)
[2016/09/07 19:14:00 | TaskManager | Error] There is 2.0 cores and 2g system memory available.
```

Second commit error message:

```
java.lang.IllegalArgumentException: <redacted> command line validation error: Heap size 1g (-Xmx) must less than the total memory of 1024k.
```

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
@@ -48,6 +48,7 @@ object SystemResources {
   def apply(cores: Option[Cores] = None, totalMemory: Option[Memory] = None) : SystemResources = {
     val heapSize = Resource.heapSize
     val system   = totalMemory match {
+      case Some(memory) if heapSize >= memory => throw new IllegalArgumentException(s"Heap size ${heapSize.prettyString} (-Xmx) must less than the total memory of ${memory.prettyString}.")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -48,6 +48,7 @@ object SystemResources {
   def apply(cores: Option[Cores] = None, totalMemory: Option[Memory] = None) : SystemResources = {
     val heapSize = Resource.heapSize
     val system   = totalMemory match {
+      case Some(memory) if heapSize >= memory => throw new IllegalArgumentException(s"Heap size ${heapSize.prettyString} (-Xmx) must greater than or equal to the total memory of ${memory.prettyString}.")
```
tfenne: The message here seems backwards.  You're saying if the `heapSize >= memory` it's bad, but the message says it must be greater than the total memory.  Presumably you mean less...

tfenne:
Nice.  One comment about > vs < then merge.

PR #217, opened 2016-09-13
Switch logging to stderr instead of stdout.

nh13:


nh13:
@tfenne I will merge this given your thumbs up on the same in `fgbio` when the tests pass.

PR #218, opened 2016-09-14
Expose an option to create the BAM index in GatherBamFiles.

nh13:
This could be set to false when gathering unsorted BAMs.


tfenne:
Looks good.  Merge at will.

PR #219, opened 2016-09-14
A couple of minor tweaks to tasks.

tfenne:


tfenne:
5 lines of changes here total @nh13

tasks/src/main/scala/dagr/tasks/bwa/Bwa.scala
```diff
@@ -61,7 +61,7 @@ object Bwa extends Configuration {
     val bwaBuffer  = new FifoBuffer[Sam].requires(memory=fifoMem)
     val altPipe    = if (doAlt) new BwaK8AltProcessor(altFile=alt) | new FifoBuffer[Sam].requires(memory=fifoMem) else Pipes.empty[Sam]
     val mergeBam   = new MergeBamAlignment(unmapped=unmappedBam, mapped=Io.StdIn, out=mappedBam, ref=ref, sortOrder=sortOrder, orientation=orientation)
-    mergeBam.requires(memory=Memory(mergeBamAlignmentMem))
+    mergeBam.requires(cores=Cores(0), memory=Memory(mergeBamAlignmentMem))
```
nh13, line 64: Why zero cores for MergeBamAlignment?  I don't see the same for SamToFastq or the FifoBuffers.

tasks/src/main/scala/dagr/tasks/fgbio/CallMolecularConsensusReads.scala
```diff
@@ -36,12 +36,13 @@ import scala.collection.mutable.ListBuffer
  */
 class CallMolecularConsensusReads(val in: PathToBam,
                                   val out: PathToBam,
-                                  val rejects: PathToBam,
+                                  val rejects: Option[PathToBam] = None,
```
nh13, line 39: good catch

tasks/src/main/scala/dagr/tasks/fgbio/CallMolecularConsensusReads.scala
```diff
@@ -36,12 +36,13 @@ import scala.collection.mutable.ListBuffer
...
                                   val tag: Option[String] = None,
                                   val readNamePrefix: Option[String] = None,
                                   val readGroupId: Option[String] = None,
                                   val errorRatePreUmi: Option[Int] = None,
                                   val errorRatePostUmi: Option[Int] = None,
+                                  val minInputBaseQuality: Option[Int] = None,
```
nh13, line 45: thanks!

nh13 (commented):
I am curious about the zero cores for `MergeBamAlignment`, after that is satisfied, 👍 .

tasks/src/main/scala/dagr/tasks/bwa/Bwa.scala
tfenne: I did this a while ago, so it's a little fuzzy.  Mostly I felt like mergeBam is doing very little most of the time while bwa is running, so allocating a core to it caused bwa to have to run with fewer cores.  I guess I'd rather over-subscribe a little bit here rather than under-subscribe.  I can flip it back if you'd rather.

tasks/src/main/scala/dagr/tasks/fgbio/CallMolecularConsensusReads.scala
tfenne: I think I changed the tool in fgbio too.  I was getting tired of writing enormous rejects.bam files that I never looked at ;)

tasks/src/main/scala/dagr/tasks/bwa/Bwa.scala
nh13: 👍

nh13 (approved):

PR #220, opened 2016-09-15
Fix to VarDict where SortVcf was failing intermittently due to streaming.

tfenne:
Also new tasks for making directories and running CollectSequencingArtifactMetrics

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/picard/CollectSequencingArtifactMetrics.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/CollectSequencingArtifactMetrics.scala
@@ -1,0 +1,58 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package dagr.tasks.picard
+
+import dagr.tasks.DagrDef._
+
+import scala.collection.mutable.ListBuffer
+
+class CollectSequencingArtifactMetrics(val in: PathToBam,
+                                       override val prefix: Option[PathPrefix] = None,
+                                       val ref: PathToFasta,
+                                       val intervals: Option[PathToIntervals] = None,
+                                       val variants: Option[PathToVcf] = None,
+                                       val minBq: Option[Int] = None,
+                                       val minMq: Option[Int] = None,
+                                       val includeUnpaired: Option[Boolean] = None,
+                                       val contextSize: Option[Int] = None,
+                                       val extension: Option[String] = Some(".txt'")
+                                      ) extends PicardTask with PicardMetricsTask {
+
+  // No single extension since multiple outputs
+  override def metricsExtension: String = ""
+
+  override protected def addPicardArgs(buffer: ListBuffer[Any]): Unit = {
+    buffer += s"I=$in"
+    buffer += s"O=$pathPrefix"
+    buffer += s"R=$ref"
+    intervals.foreach      (i => buffer += s"INTERVALS=$i")
+    variants.foreach       (v => buffer += s"DB_SNP=$v")
+    minBq.foreach          (q => buffer += s"Q=$q")
+    minMq.foreach          (q => buffer += s"MQ=$q")
+    includeUnpaired.foreach(i => buffer += s"INCLUDE_UNPAIRED=$i")
+    contextSize.foreach    (s => buffer += s"CONTEXT_SIZE=$s")
+    extension.foreach      (e => buffer += s"EXT=$e")
+  }
+}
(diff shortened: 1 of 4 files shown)


tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -50,13 +51,13 @@ object VarDictJava extends Configuration {
   val RootDirConfigKey = "vardictjava.dir"
 
   /** Path to the VarDictJava directory. */
-  lazy val RootDir = configure[Path](RootDirConfigKey)
+  val RootDir = configure[Path](RootDirConfigKey)
```
tfenne, line 54: I made these non-lazy so that any configuration issues will show up at the first reference to the class/object instead of much later when the tasks try to run.  I find it much more helpful for debugging new setups.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -123,12 +122,14 @@ class VarDictJavaEndToEnd(tumorBam: PathToBam,
                           maxThreads: Int = 32) extends Pipeline {
   import VarDictJava.BinDir
 
+  // Put these here so that they'll error at construction if missing
```
tfenne, line 125: Same thing here - moving them up to the constructor surfaces any problems sooner.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -123,12 +122,14 @@ class VarDictJavaEndToEnd(tumorBam: PathToBam,
...
+  private val biasScript = BinDir.resolve("teststrandbias.R")
+  private val vcfScript  = BinDir.resolve("var2vcf_valid.pl")
+
   def build(): Unit = {
     val tn = tumorName.getOrElse(VarDictJava.extractSampleName(bam=tumorBam))
     val dict = PathUtil.replaceExtension(ref, ".dict")
-
-    val biasScript = BinDir.resolve("teststrandbias.R")
-    val vcfScript  = BinDir.resolve("var2vcf_valid.pl")
+    val tmpVcf = Io.makeTempFile("vardict.", ".vcf", dir= if (out == Io.StdOut) None else Some(out.getParent))
```
tfenne, line 132: This is a little hokey, but without having the constructor take in a new directory arg it's not clear what would be better.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -145,11 +146,11 @@ class VarDictJavaEndToEnd(tumorBam: PathToBam,
...
     val toVcf              = new ShellCommand(vcfScript.toString, "-N", tn, "-E", "-f", minimumAf.toString) with Pipe[Any,Vcf]
     val removeRefEqAltRows = new ShellCommand("awk", "{if ($1 ~ /^#/) print; else if ($4 != $5) print}") with Pipe[Vcf,Vcf]
-    val sortVcf            = new SortVcf(dict=Some(dict))
+    val sortVcf            = new SortVcf(in=tmpVcf, out=out, dict=Some(dict))
 
     bias.requires(Cores(0), Memory("32m"))
     toVcf.requires(Cores(0), Memory("32m"))
 
-    root ==> (vardict | bias | toVcf | removeRefEqAltRows | sortVcf > out).withName(this.getClass.getSimpleName)
+    root ==> (vardict | bias | toVcf | removeRefEqAltRows > tmpVcf).withName("VarDictJava") ==> sortVcf ==> new DeleteFiles(tmpVcf)
```
tfenne, line 154: I did this for two reasons:
1. I was getting failures with streaming into SortVcf.  But the exact same commands run with writing to an intermediate file worked just fine.
2. The way the SortVcf task works, if it's output is `Io.Stdout` it won't create an index for you, and I want an index.

tfenne (commented):
@nh13 I suspect you won't love some of the changes to VarDictJava but they were necessary to get it working for me.

tasks/src/main/scala/dagr/tasks/picard/CollectSequencingArtifactMetrics.scala
```diff
@@ -0,0 +1,56 @@
...
+class CollectSequencingArtifactMetrics(val in: PathToBam,
+                                       override val prefix: Option[PathPrefix] = None,
+                                       val ref: PathToFasta,
+                                       val intervals: Option[PathToIntervals] = None,
+                                       val variants: Option[PathToVcf] = None,
+                                       val minBq: Option[Int] = None,
+                                       val minMq: Option[Int] = None,
+                                       val contextSize: Option[Int] = None,
+                                       val extension: Option[String] = Some(".txt'")
+                                      ) extends PicardTask with PicardMetricsTask {
```
nh13, line 41: Can you add support for the `INCLUDE_UNPAIRED` argument?  I am likely going to need it if this task is in dagr.

tasks/src/main/scala/dagr/tasks/picard/CollectSequencingArtifactMetrics.scala
```diff
@@ -0,0 +1,56 @@
...
+  override protected def addPicardArgs(buffer: ListBuffer[Any]): Unit = {
+    buffer += s"I=$in"
+    buffer += s"O=$pathPrefix"
+    buffer += s"R=$ref"
+    intervals.foreach  (i => buffer += s"INTERVALS=$i")
+    variants.foreach   (v => buffer += s"DB_SNP=$v")
+    minBq.foreach      (q => buffer += s"Q=$q")
+    minMq.foreach      (q => buffer += s"MQ=$q")
+    contextSize.foreach(s => buffer += s"CONTEXT_SIZE=$s")
+    extension.foreach  (e => buffer += s"EXT=$e")
```
nh13: Any reason why you are not using `intervals.foreach.foreach(i => buffer.append("INTERVALS=" + i))`?  I am looking at other tasks in this package and seeing both `+=` and `append` used, but rarely `s"...=$i"`.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: 👍 as long as it doesn't break when I don't use `VarDictJava` in a pipeline.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: (1) Thanks `SortVcf`.  I am curious about the error and if that is something we should log an issue in Picard.
(2) This confuses me: if the final output is `/dev/stdout` then how could you want an index?

nh13 (commented):
Some minor stuff, please respond and then 👍 merge.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
tfenne: On (2) - it was that even when I set `out=someVcf` on the pipeline, the pipeline was leaving it as `Io.Stdout` on `SortVcf` and using a `>` to push it into the file.  So I'm writing to a file, but `SortVcf` thinks it's writing to stdout.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
tfenne: Yup - it should only matter if a pipeline references the object, or constructs an instance of the class.

tasks/src/main/scala/dagr/tasks/picard/CollectSequencingArtifactMetrics.scala
tfenne: LOL, mostly because I feel like you prefer the `s""` syntax most places to construct strings. I don't really care all that much either way.

PR #221, opened 2016-09-17
If we attempt to terminate a task that has succeeded, mark it as succ…

nh13:
…eeded not stopped.

Consider the case we have a failed tasks, so the `TaskManager` terminates all running tasks, but some of those running tasks have actually completed successfully.  Previously they would have status `STOPPED` while they should be `SUCCEEDED`.  Furthermore, running tasks that were terminated (interrupted) had status `FAILED_COMMAND` rather than `STOPPED`.  This kills two birds with one stone.

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/TaskExecutionRunner.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskExecutionRunner.scala
@@ -144,7 +144,7 @@ private object TaskExecutionRunner {
   }
 }
 
-/** Tracks and run a set of [[Task]]s. */
+/** Tracks and run a set of [[dagr.core.tasksystem.Task]]s. */
 private[core] class TaskExecutionRunner extends TaskExecutionRunnerApi with LazyLogging {
   import TaskExecutionRunner._
   private val processes: mutable.Map[TaskId, Thread] = new mutable.HashMap[TaskId, Thread]()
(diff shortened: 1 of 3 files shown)


core/src/main/scala/dagr/core/execsystem/TaskExecutionRunner.scala
```diff
@@ -205,10 +205,13 @@ private[core] class TaskExecutionRunner extends TaskExecutionRunnerApi with Lazy
 
 
     taskInfo.endDate = Some(Instant.now())
```
tfenne: Should this also be moved inside the new `if`?

core/src/main/scala/dagr/core/execsystem/TaskExecutionRunner.scala
```diff
@@ -253,18 +256,18 @@ private[core] class TaskExecutionRunner extends TaskExecutionRunnerApi with Lazy
   override def terminateTask(taskId: TaskId): Boolean = {
       processes.get(taskId) match {
         case Some(thread) =>
+          val taskInfo = taskInfos.get(taskId) match {
```
tfenne: If this is truly exceptional, why not just write `taskInfos(taskid)`?

tfenne (approved):
A couple of minor comments, otherwise looks good.

PR #225, opened 2016-09-28
Added in caching of type -> class conversion.

tfenne:
@nh13 This makes a small difference in dagr, but a huge difference in `fgbio`'s delimited parsing.  Dropped my time for parsing a really large file from ~30 seconds to < 2 seconds!


nh13:
@tfenne likely it is failing due to a travis-ci bug: https://github.com/travis-ci/travis-ci/issues/6641

nh13 (approved):

PR #227, opened 2016-10-07
Do not merge: Expose a few command line options for scheduling strategies.

nh13:
@tfenne looking for some feedback.  My only regret is having to use a Java enum to expose the strategy options on the command line.  Perhaps I could play with `ClassFinder` to find all classes that extend `SelectTaskScheduler` instead.  Tests will be added and code cleaned up when the idea becomes more fully formed.


nh13:
Bump.  Given a few more multi-sample pipelines and whole genome pipelines, this may be more relevant. @tfenne thoughts?

PR #229, opened 2016-10-07
Added the ref parameter to TrimPrimers.

tfenne:
@nh13: A two-liner PR.  Note: I left the ref without a default `= None` since I think it's probably better to prompt users (i.e. me) to make a decision when using it.


nh13:
image

PR #231, opened 2016-11-07
Set compression in an Fgbio task.

nh13:


tasks/src/main/scala/dagr/tasks/fgbio/FgBioTask.scala
```diff
@@ -27,10 +27,11 @@ package dagr.tasks.fgbio
 import java.nio.file.Path
 
 import dagr.core.config.Configuration
-import dagr.core.execsystem.{Memory, Cores}
+import dagr.core.execsystem.{Cores, Memory}
```
tfenne, line 30: LOL.  Making the important changes 😉

tasks/src/main/scala/dagr/tasks/fgbio/FgBioTask.scala
```diff
@@ -39,8 +40,11 @@ object FgBioTask {
 
 /**
   * Base Task for any task in the FgBio jar.
+  *
+  * @param compressionLevel the compress level to use for HTSJDK.
   */
-abstract class FgBioTask extends ProcessTask with JarTask with FixedResources with Configuration {
+abstract class FgBioTask(var compressionLevel: Option[Int] = None)
```
tfenne, line 46: I wonder if this should be `bamCompressionLevel` or something more specific?

tfenne (approved):

PR #233, opened 2017-01-02
MarkDuplicates should handle stdin.

nh13:
If the input is stdin, then we try and use the output file to infer the name of the metrics file.


tfenne:
Seems fine to me ... though I thought MarkDuplicates couldn't handle `stdin` since it reads through it's inputs twice?

nh13:
Yes, `MarkDuplicates` sadly does not support streaming.  Closing.

PR #241, opened 2017-01-16
Add options to MergeBamAlignment task for rev'ing and rc'ing tags.

tfenne:
Shouldn't be merged until https://github.com/broadinstitute/picard/pull/731 is merged.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/picard/MergeBamAlignment.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/MergeBamAlignment.scala
@@ -42,11 +42,13 @@ class MergeBamAlignment(unmapped: PathToBam,
                         programGroupCommand: Option[String] = None,
                         programGroupName: Option[String] = None,
                         attributesToRetain: List[String] = List[String]("X0", "ZS", "ZI", "ZM", "ZC", "ZN"),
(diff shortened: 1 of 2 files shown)


tasks/src/main/scala/dagr/tasks/picard/MergeBamAlignment.scala
```diff
@@ -42,11 +42,13 @@ class MergeBamAlignment(unmapped: PathToBam,
                         programGroupCommand: Option[String] = None,
                         programGroupName: Option[String] = None,
                         attributesToRetain: List[String] = List[String]("X0", "ZS", "ZI", "ZM", "ZC", "ZN"),
+                        attributesToReverse: List[String] = List[String]("ad", "bd", "cd", "ae", "be", "ce"), // fgbio consensus tags
```
nh13, line 45: IT would be nice to put these in an `Fgbio` object or similar in case we want to reference them in another task.  I just don't want to forget about them here.

nh13:
One minor comment, take it or leave it, then 👍

tasks/src/main/scala/dagr/tasks/picard/MergeBamAlignment.scala
tfenne: Once we get the duplex caller done they'll all be in the `ConsensusTags` object.  Just not quite there yet and didn't want to have to make a second PR here later.

PR #243, opened 2017-01-17
Estimate the resources needed for a VariableResources task.

nh13:
This is used when a VariableResources task cannot be scheduled, and we'd like to know the minimum amount of resources it needs.

@tfenne I should add some tests, but I wanted to see what you thought.  I could also try a search on the diagonal too, but I thought that was overkill.

### DIFF
--- a/core/src/main/scala/dagr/core/tasksystem/Schedulable.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/Schedulable.scala
@@ -23,13 +23,55 @@
  */
 package dagr.core.tasksystem
 
-import dagr.core.execsystem.{Cores, Memory, ResourceSet}
+import dagr.core.execsystem.{Cores, Memory, Resource, ResourceSet}
+
+object Schedulable {
+
+  /** The multiple of the system resources to use when searching for the resources with which a task can execute. */
+  private val ResourceMultiple: Int = 16
+
+  /** The maximum # cores to use when searching for the resources with which a task can execute. */
+  private[tasksystem] val EndCores: Cores = {
+    val cores    = Resource.systemCores.value * ResourceMultiple
+    val minCores = 64
+    if (cores < minCores) Cores(minCores) else Cores(cores)
+  }
+
+  /** The maximum amount of memory to use when searching for the resources with which a task can execute. */
+  private[tasksystem] val EndMemory: Memory = {
+    val memory    = Resource.systemMemory.value * ResourceMultiple
+    val minMemory = Memory("256g")
+    if (memory < minMemory.value) minMemory else Memory(memory)
+  }
+
+  /** The maximum resources to use when searching for the resources with which a task can execute. */
+  private[tasksystem] val EndResources: ResourceSet = new ResourceSet(cores=EndCores, memory=EndMemory)
+
+  /** The amount of memory per core when searching for the memory with which a task can execute. */
+  private[tasksystem] val MemoryPerCore = Memory("1g")
+
+  /** The minimum amount of memory to start when searching for the memory with which a task can execute. */
+  private[tasksystem] val StartMemory = Memory("256m")
+
+  /** The minimum amount of memory to increase the memory when searching for the memory with which a task can execute. */
+  private[tasksystem] val StepMemory: Memory = {
+    val stepMemory = Memory(EndMemory.value / 100)
+    if (stepMemory < StartMemory) stepMemory
+    else StartMemory
+  }
+
+  /** A single core! */
+  private[tasksystem] val OneCore: Cores = Cores(1)
+}
 
 /**
   * Traits that isolates methods about how Tasks interact with the Scheduler, and allows
   * for multiple implementations independent from the Task hierarchy.
   */
 trait Schedulable {
+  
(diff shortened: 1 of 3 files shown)


core/src/main/scala/dagr/core/tasksystem/Schedulable.scala
```diff
@@ -89,4 +89,30 @@ trait FixedResources extends ScheduleWithEmptyDefaultResources {
...
+  }
+
+  /** A crude attempt at estimating the minimum amount of resources needed by this task.  First estimates the # of cores
+    * by assuming an infinite amount of memory, and then estimates the amount of memory assuming an infinite # of cores.
+    * If either does not yield a value, an infinite amount for that resource type is used.
+    */
+  def minCoresAndMemory(maxCores: Cores = Cores(32), maxMemory: Memory = Memory("64g")): ResourceSet = {
+    val cores  = minCores(end=maxCores).getOrElse(Cores.infinite)
+    val memory = minMemory(end=maxMemory).getOrElse(Memory.infinite)
+    new ResourceSet(cores, memory)
```
tfenne: Perhaps at least test that if you offer it `cores` and `memory` after computing them separately, that that's sufficient?

tfenne (changes requested):
I think I would do a couple of things in addition to this:
1) Perhaps before searching try offering the `end` amounts of cores and memory together and make sure that's actually sufficient in combination
2) Once you have the mins for both, I'd suggest re-checking that the combination is acceptable, and if not, return infinite or change to returning `Option[ResourceSet]`.

nh13:
@tfenne made some changes you suggested and more.  I added in the ability to search using per-core memory ratio, which is common across my pipelines.  I moved things up into `Schedulable` too.  Could you take another look?

core/src/main/scala/dagr/core/tasksystem/Schedulable.scala
```diff
@@ -48,6 +48,65 @@ trait Schedulable {
...
+    * @param maximumResources The maximum system resources available to the task.
+    * @return Either a ResourceSet of the subset of resources that this task can run with, or None
+    */
+  private[core] def minResources(maximumResources: ResourceSet = new ResourceSet(cores=Cores(32), memory=Memory("64g"))): Option[ResourceSet] = {
+    val cores  = minCores(end=maximumResources.cores, maxMemory=maximumResources.memory)
+    val memory = minMemory(end=maximumResources.memory, maxCores=maximumResources.cores)
+    (cores, memory) match {
+      case (Some(c), Some(m)) => this.pickResources(maximumResources.copy(cores=c, memory=m))
+      case _                  =>
+        // try a combination of cores and memory with a fixed reatio between the two.
```
tfenne: reatio -> ratio

core/src/main/scala/dagr/core/tasksystem/Schedulable.scala
```diff
@@ -48,6 +48,65 @@ trait Schedulable {
...
+    *
+    * First estimates the # of cores by assuming an infinite amount of memory, and then estimates
+    * the amount of memory assuming an infinite # of cores.  If either does not yield a value or if
+    * the task cannot run with the combination of the two, then try a various combinations of cores
+    * and memory where we assume a fixed ratio between cores and memory (ex. 1g per core).
+    *
+    * @param maximumResources The maximum system resources available to the task.
+    * @return Either a ResourceSet of the subset of resources that this task can run with, or None
+    */
+  private[core] def minResources(maximumResources: ResourceSet = new ResourceSet(cores=Cores(32), memory=Memory("64g"))): Option[ResourceSet] = {
```
tfenne: The defaults here feel very "no-one will ever need more than 640kb of memory"-esque.  I would either set these way higher (like 1PB and 1000 cores) or set the as some very large multiple of the current system resources.

core/src/main/scala/dagr/core/tasksystem/Schedulable.scala
```diff
@@ -48,6 +48,65 @@ trait Schedulable {
...
+    val cores  = minCores(end=maximumResources.cores, maxMemory=maximumResources.memory)
+    val memory = minMemory(end=maximumResources.memory, maxCores=maximumResources.cores)
+    (cores, memory) match {
+      case (Some(c), Some(m)) => this.pickResources(maximumResources.copy(cores=c, memory=m))
+      case _                  =>
+        // try a combination of cores and memory with a fixed reatio between the two.
+        val startMemoryPerCore = Memory("256m").value
+        val endMemoryPerCore   = maximumResources.memory.value
+        val stepMemoryPerCore  = Memory("256m").value
```
tfenne: Instead of hard-coding this to 256m, could you set it to be some reasonable fraction (e.g. 1/100th) of the max memory so you have a fixed number of steps?

tfenne (commented):
Sorry @nh13 I somehow missed that this was re-assigned to me for another look.  Some very minor suggestions (mostly around upping the max values) and then merge when you're ready.

nh13:
@tfenne I made the search based on a multiple of the system resources.  If we find that these estimates don't yield good results, we can always improve them.  Lets table that discussion since I have a lot of complexity in their for now.

tfenne (approved):

tfenne:
Failing tests need fixing, then 👍

PR #246, opened 2017-01-20
Small tweaks to how collection options & default values are output.

tfenne:
@nh13 Small change because it was bothering me how Collection/Traversable arguments were output in the usage.  Two changes:

1. The more important one, to how default values are output, so it drops the `List()` or `Set()` wrapper and outputs `A, B, C` instead of `List(A, B, C)`.
2. Outputting the range in a way I think is more understandable to those who aren't intimately familiar with regex details.  So `Int{1, 3}` is not `Int{1-3}`.  I'd also be happy with `Int{1..3}` if you thought that was clearer.

I was also tempted to write it as `Int[1-3]` using square brackets, but thought that might be confusing given the use of square brackets for optional values.

### DIFF
--- a/sopt/src/main/scala/dagr/sopt/cmdline/ClpArgumentDefinitionPrinting.scala
+++ b/sopt/src/main/scala/dagr/sopt/cmdline/ClpArgumentDefinitionPrinting.scala
@@ -24,11 +24,13 @@
 
 package dagr.sopt.cmdline
 
+import java.util
+
 import dagr.commons.reflect.ReflectionUtil
 import dagr.sopt.util._
 import dagr.commons.util.StringUtil
 
-import scala.util.{Success, Failure}
+import scala.util.{Failure, Success}
 
 object ClpArgumentDefinitionPrinting {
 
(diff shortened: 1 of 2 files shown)


nh13:
1. Can `List(A, B, C)` be `[A, B, C]` or `{A, B, C}`?  Looking at just `A, B, C` doesn't emphasize that its a sequence of values.
2.  I'd prefer `Int{1..3}` but that's my 2 cents.

sopt/src/main/scala/dagr/sopt/cmdline/ClpArgumentDefinitionPrinting.scala
```diff
@@ -80,7 +82,7 @@ object ClpArgumentDefinitionPrinting {
     val desciption = (argumentDefinition.minElements, argumentDefinition.maxElements) match {
       case (0, Integer.MAX_VALUE) => "*"
       case (1, Integer.MAX_VALUE) => "+"
-      case (m, n)                 => s"{$m, $n}"
+      case (m, n)                 => s"{${m}-${n}}"
```
nh13: This probably needs a test ;)

tfenne:
@nh13 I've changed to use `1..3` instead of `1-3`.  I'm hesitant to put brackets around the list because there are already one set of brackets from the formatting.  I.e. I have it changed from:
```
-M Int{1, 3}, --min=Int{1, 3}      The minimum number of things.  [Default: List(1,2,3)].
```
to
```
-M Int{1..3}, --min=Int{1..3}      The minimum number of things.  [Default: 1, 2, 3].
```
because I think this looks funny:
```
-M Int{1..3}, --min=Int{1..3}      The minimum number of things.  [Default: [1, 2, 3]].
```
If you still think the last one looks best in context I'll add more brackets.

nh13:
👍

PR #248, opened 2017-01-22
Reving library and plugin dependencies to stay up to date.

tfenne:
Some of our plugin dependencies were old enough that they're not built for 2.12.  With the updates one can switch the scala version in the build file back and forth between 2.12 and 2.11 and everything compiles and tests pass.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,15 +1,9 @@
 resolvers += Resolver.url("fix-sbt-plugin-releases", url("http://dl.bintray.com/sbt/sbt-plugin-releases"))(Resolver.ivyStylePatterns)
 
-addSbtPlugin("com.typesafe.sbt" % "sbt-git" % "0.8.5")
-
-addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.0")
-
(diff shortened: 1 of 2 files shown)


project/plugins.sbt
```diff
@@ -1,15 +1,9 @@
...
-
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.0.1")
-
-addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "1.1")
-
-addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
-
-addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.3.3")
+addSbtPlugin("com.typesafe.sbt"  % "sbt-git"       % "0.8.5")
+addSbtPlugin("com.github.gseitz" % "sbt-release"   % "1.0.3")
```
nh13, line 4: My spider senses are saying that there was some issue with selecting the specific versions for two or more of these dependencies.  I suppose we wont find out until the next release, but it's just a head's up.  I thought I would have added a comment somewhere here, but I didn't, so who's to say.

tfenne: Well, if we run into issues at the next release it should be easy enough to restore the old versions.

PR #249, opened 2017-01-23
Addition of CallDuplexConsensusReads task and updates to...

tfenne:
...CallMolecularConsensusReads and FilterConsensusReads.

Relies on the changes in https://github.com/fulcrumgenomics/fgbio/pull/159 and should not be merged before that PR.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/fgbio/CallDuplexConsensusReads.scala
+++ b/tasks/src/main/scala/dagr/tasks/fgbio/CallDuplexConsensusReads.scala
@@ -1,0 +1,51 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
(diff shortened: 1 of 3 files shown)


tasks/src/main/scala/dagr/tasks/fgbio/CallDuplexConsensusReads.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import scala.collection.mutable.ListBuffer
+
+class CallDuplexConsensusReads(val in: PathToBam,
+                               val out: PathToBam,
+                               val readNamePrefix: Option[String] = None,
+                               val readGroupId: Option[String] = None,
+                               val errorRatePreUmi: Option[Int] = None,
+                               val errorRatePostUmi: Option[Int] = None,
+                               val minInputBaseQuality: Option[Int] = None
```
nh13: The rule about when to align the equals escapes me.

tasks/src/main/scala/dagr/tasks/fgbio/FilterConsensusReads.scala
```diff
@@ -29,26 +29,38 @@ import dagr.tasks.DagrDef.PathToBam
...
+            maxReadErrorRate: Double,
+            maxBaseErrorRate: Double,
+            minQuality: Int,
+            maxNoCallFraction: Double): FilterConsensusReads = {
+    apply(in=in, out=out, ref=ref, minReads=Seq(minReads), maxReadErrorRate=Seq(maxReadErrorRate),
+      maxBaseErrorRate=Seq(maxBaseErrorRate), minQuality=minQuality, maxNoCallFraction=maxNoCallFraction)
+  }
+}
+
+case class FilterConsensusReads(in: PathToBam,
```
nh13, line 46: I assume this is a case class now for pipelines that want to iterate over parameter lists and therefore use the `copy` method?

tasks/src/main/scala/dagr/tasks/fgbio/FilterConsensusReads.scala
```diff
@@ -29,26 +29,38 @@ import dagr.tasks.DagrDef.PathToBam
...
+                                out: PathToBam,
+                                ref: PathToFasta,
+                                minReads: Seq[Int],
+                                maxReadErrorRate: Seq[Double],
+                                maxBaseErrorRate: Seq[Double],
```
nh13, line 51: I am torn if these should have defaults (i.e. `Seq.emtpy`).

nh13 (approved):
A penny for your thoughts on some comments, otherwise it looks good.

tasks/src/main/scala/dagr/tasks/fgbio/CallDuplexConsensusReads.scala
tfenne: Lol.  I'll align these.  I don't think we have a _rule_ per se.  I just tend to align things when it bothers me 😛

tasks/src/main/scala/dagr/tasks/fgbio/FilterConsensusReads.scala
tfenne: That and also because I wanted to have a pair of apply methods, one that took Seqs and one that took single values, because I think that's nicer than having two constructors, and this way I get one of them for free.

tasks/src/main/scala/dagr/tasks/fgbio/FilterConsensusReads.scala
tfenne: The `fgbio` tool requires at least one value for each of these arguments.  I.e. to turn off the filtering you set it to `1.0`, but you don't omit it.  So defaulting to `Seq.empty` here would be misleading I think.

PR #251, opened 2017-01-24
Add a message when execution completes.

nh13:


tfenne (commented):
What is the output if I'm an idiot and have orphaned tasks?  Otherwise, looks good to me.

PR #252, opened 2017-01-26
The ClipOverlappingReads task retries with more memory

nh13:


nh13:
Closing as I will mix it in myself.

PR #253, opened 2017-01-30
Expose VarDictJava on the command line.

nh13:
Also fixed a bug where VarDictJava would require one more core than needed.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
+++ b/tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
@@ -29,13 +29,16 @@ import java.nio.file.Path
 
 import dagr.commons.CommonsDef.yieldAndThen
 import dagr.commons.io.{Io, PathUtil}
+import dagr.core.cmdline.Pipelines
 import dagr.core.config.Configuration
 import dagr.core.execsystem.{Cores, Memory, ResourceSet}
+import dagr.core.tasksystem.Pipes.PipeWithNoResources
 import dagr.core.tasksystem._
-import dagr.tasks.DagrDef.{FilePath, PathToBam, PathToFasta, PathToVcf}
+import dagr.sopt.{arg, clp}
+import dagr.tasks.DagrDef._
 import dagr.tasks.DataTypes.Vcf
 import dagr.tasks.misc.DeleteFiles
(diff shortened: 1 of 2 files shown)


tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -108,24 +111,32 @@ private class VarDictJava(tumorBam: PathToBam,
...
+      |Tumor-only single-sample variant calling with VarDictJava: https://github.com/AstraZeneca-NGS/VarDictJava
+      |
+      |The output Tumor name will be inferred from the the first read group in the input BAM if not provided.
+    """",
+  group = classOf[Pipelines]
+)
+class VarDictJavaEndToEnd
+(@arg(flag="i", doc="The input tumor BAM file.") tumorBam: PathToBam,
+ @arg(flag="l", doc="The intervals oer which to call variants.") intervals: PathToIntervals,
+ @arg(flag="r", doc="Path to the reference fasta file.")         val ref: PathToFasta,
```
tfenne: Extra spaces.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -108,24 +111,32 @@ private class VarDictJava(tumorBam: PathToBam,
...
+      |
+      |Tumor-only single-sample variant calling with VarDictJava: https://github.com/AstraZeneca-NGS/VarDictJava
+      |
+      |The output Tumor name will be inferred from the the first read group in the input BAM if not provided.
+    """",
+  group = classOf[Pipelines]
+)
+class VarDictJavaEndToEnd
+(@arg(flag="i", doc="The input tumor BAM file.") tumorBam: PathToBam,
+ @arg(flag="l", doc="The intervals oer which to call variants.") intervals: PathToIntervals,
```
tfenne: oer -> over

tfenne (commented):
The only thing I'm not 100% keen on is the change to take in an interval list and convert it to a BED.  Mostly because I use this in a lot of pipelines, and I like to keep the BED file around for use after the fact.

nh13:
I kept the `bed` option and addressed your comments, merging once it passes.

PR #256, opened 2017-02-07
Added ability to use triple-quoted, margined strings for argument docs.

tfenne:
@nh13 the scaladoc process doesn't like @arg annotations with `doc="foo" + "bar"` which is the obvious way to type out longer docs.  This change allows `doc=` to use the same format as we do for `@clp(description)` but removes line breaks and replaces them with spaces.


nh13:
How about a test just to make sure this doesn't get reverted?

PR #258, opened 2017-02-11
Made the assignTag optional in GroupReadsByUmi since it has a default in the CLP

tfenne:
Another trivial PR @nh13 - I'm not sure why we made this required in the first place since the CLP defaults to `MI`.  I don't want to have to repeat that everywhere!


nh13:
I have pipelines that will break with this change :/  The only alternative is to set the default value to `MI`.

tfenne:
Me too, but it's a fairly simple change to make when updating.  If you object I'll just default it to `MI`, but I think this is the better change.

nh13:
It's fine just fine.

PR #259, opened 2017-02-11
Dependency DSL now works on Options!

tfenne:
@nh13 This PR is a touch more involved, but I'd love to get your feedback.  I got tired of having to do things like:

```scala
if (x) new Task(...) else NoOpInJvmTask(...)
```

in order to keep the dependency DSL happy.  The goal here is to allow `Option[Dependable]` to function as a dependable.  `None`s are effectively flattened out - i.e. `x ==> None == y` is flattened to `x ==> y` and `x :: None :: y` is flattened to `x :: y`.

This allows a much more natural approach like:

```scala
val makeFoo = foo.map(f => new MakeFoo(in=x, out=f))
val qcFoo = foo.map(f => new QcFoo(in=f, out=qc))
root ==> makeFoo ==> qcFoo
```

The implicit is only required to make things work when an `Option[Dependable]` shows up at the end of a chain (either lhs for `==>` or rhs for `::` due to associativity.  I.e.

```scala
// Works without implicit
root ==> None ==> Some(task) ==> None
root ==> (None :: Some(task) :: otherTask)

// Works only with the implicit
Some(task) ==> otherTask
root ==> (task :: Some(otherTask))
```

### DIFF
--- a/core/src/main/scala/dagr/core/tasksystem/Dependable.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/Dependable.scala
@@ -68,7 +68,7 @@ trait Dependable {
 
 /** An object that can be implicitly converted to from a None when using Option[Dependable]. */
 object EmptyDependable extends Dependable {
-  /** Converts an Option[Dependable] to a Dependable when needed. */
+  /** Converts an Option[Dependable] to a Dependable when needed. Linked to from [[DagrDef]] to accessibility. */
   implicit def optionDependableToDependable(maybe: Option[Dependable]): Dependable = maybe match {
     case Some(d) => d
     case None    => EmptyDependable
--- a/core/src/main/scala/dagr/core/tasksystem/Dependable.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/Dependable.scala
@@ -29,9 +29,18 @@ package dagr.core.tasksystem
   */
 trait Dependable {
   /** Creates a dependency on this dependable, for the provided Task. */
-  final def ==> (other: Dependable) : Dependable = {
-    addDependent(other)
-    new DependencyChain(this, other)
+  final def ==> (other: Dependable) : Dependable = (this, other) match {
+    case (EmptyDependable, d) => d
+    case (d, EmptyDependable) => d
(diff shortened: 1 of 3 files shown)


tfenne:
If you're on board I'll add some more tests before merging this.

core/src/main/scala/dagr/core/tasksystem/Dependable.scala
```diff
@@ -41,17 +50,45 @@ trait Dependable {
...
+  }
 
   /** Abstract method that must be implemented to return all Tasks associated with this Dependable. */
   private[tasksystem] def toTasks: Traversable[Task]
 }
 
+/** An object that can be implicitly converted to from a None when using Option[Dependable]. */
+object EmptyDependable extends Dependable {
+  /** Converts an Option[Dependable] to a Dependable when needed. */
+  implicit def optionDependableToDependable(maybe: Option[Dependable]): Dependable = maybe match {
```
nh13: I am wondering if this is the right place for this method, since it doesn't have any outward facing references to EmptyDependable.

tasks/src/main/scala/dagr/tasks/DagrDef.scala
```diff
@@ -65,4 +67,6 @@ object DagrDef {
   /** A String that represents the prefix or basename of a filename. */
   type FilenamePrefix = String
 
+  /** Implicit that will convert an Option[Dependable] to a Dependable when needed. */
+  implicit def optionDependableToDependable(maybe: Option[Dependable]): Dependable = EmptyDependable.optionDependableToDependable(maybe)
```
nh13, line 71: Well, this too will work.

nh13:
Very cool, take it or leave it about my comment about where the implicit should go.

tasks/src/main/scala/dagr/tasks/DagrDef.scala
tfenne: Yeah, I felt like the implementation should live with all the Dependable stuff, but I wanted it to be imported into scope through `DagrDef`.

PR #261, opened 2017-02-12
Fix for minor bug where more dependencies were getting created than necessary.

tfenne:
Previously `a ==> (b ==> c ==> d)` would create `a ==> b`, `a ==> c` and `a ==> d` instead of just a ==> b.  This didn't cause any errors, but is unnecessary, and confusing when debugging.

### DIFF
--- a/core/src/main/scala/dagr/core/tasksystem/Dependable.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/Dependable.scala
@@ -62,8 +62,14 @@ trait Dependable {
     case Some(d) => this :: d
   }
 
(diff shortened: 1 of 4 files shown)


tfenne:
@nh13 Hopefully the last dagr PR this weekend.  I discovered, as I was debugging a different issue, that the dependency DSL creates a lot more dependencies than it needs to do in some circumstances.  This PR eliminates the unnecessary ones.

nh13:
LGTM although coverage goes down by a quarter percent.

tfenne:
I think the coverage is largely driven by uncovered methods in subclasses of `Dependable` which are required by the trait, but are never invoked in reality.  I'm not sure what to do about those.

PR #262, opened 2017-02-15
New Task class for FilterSomaticVcf.

tfenne:
Another tiny PR for you @nh13

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/fgbio/FilterSomaticVcf.scala
+++ b/tasks/src/main/scala/dagr/tasks/fgbio/FilterSomaticVcf.scala
@@ -1,0 +1,53 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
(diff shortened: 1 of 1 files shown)


tasks/src/main/scala/dagr/tasks/fgbio/FilterSomaticVcf.scala
```diff
@@ -0,0 +1,53 @@
...
+
+package dagr.tasks.fgbio
+
+import dagr.tasks.DagrDef.{PathToBam, PathToVcf}
+
+import scala.collection.mutable.ListBuffer
+
+case class FilterSomaticVcf(in: PathToVcf,
+                            out: PathToVcf,
+                            bam: PathToBam,
```
nh13: a minor quibble but could `bam` be before `out` since both `in` and `bam` are inputs (group together)?

nh13 (approved):
LGTM, minor quibble (it was the best I could do)

PR #269, opened 2017-03-15
Pipelines can now control printing their clp args (if they are clps).

nh13:
@tfenne I waffled on a bunch of ways and places in which to implement this, including a mix-in trait for `Pipeline` (ex. `PrintPipelineArgs`) versus the current method _in_ `Pipeline`, having the parsing code do the actual printing versus the `DagrCoreMain`.  I found this was the easiest.

### DIFF
--- a/core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
+++ b/core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
@@ -82,15 +82,15 @@ object DagrCoreMain extends Configuration {
     def mainBlock(dagr: DagrCoreMain): Unit = loadScripts(clp=dagr)
 
     // the method that should be called once the dagr and pipeline instances have been successfully created
-    def clpBlock(dagr: DagrCoreMain)(pipeline: Pipeline): Unit = dagr.configure(pipeline)
+    def clpBlock(dagr: DagrCoreMain)(pipeline: Pipeline): Unit = dagr.configure(pipeline, parser.commandLine)
 
     parser.parseCommandAndSubCommand[DagrCoreMain](
-      args                = args,
-      packageList         = packageList,
-      omitSubClassesOf    = Seq(classOf[DagrCoreMain]),
-      includeHidden       = includeHidden,
-      afterCommandBuild           = mainBlock,
-      afterSubCommandBuild            = clpBlock
+      args                 = args,
+      packageList          = packageList,
+      omitSubClassesOf     = Seq(classOf[DagrCoreMain]),
+      includeHidden        = includeHidden,
+      afterCommandBuild    = mainBlock,
+      afterSubCommandBuild = clpBlock
     )
   }
 
@@ -165,7 +165,7 @@ class DagrCoreMain(
   }
 
   // Invoked by DagrCommandLineParser after the pipeline has also been instantiated
-  private[cmdline] def configure(pipeline: Pipeline) : Unit = {
+  private[cmdline] def configure(pipeline: Pipeline, commandLine: Option[String] = None) : Unit = {
     try {
       val config = new Configuration { }
 
@@ -179,7 +179,7 @@ class DagrCoreMain(
         val errors: ListBuffer[String] = ListBuffer[String]()
         scriptsDirectory.foreach(d => mkdir(dir=d, use="scripts", errors=errors))
         logDirectory.foreach(d => mkdir(dir=d, use="logs", errors=errors))
-        if (errors.nonEmpty) throw new ValidationException(errors.toList)
+        if (errors.nonEmpty) throw ValidationException(errors.toList)
       }
 
       Logger.level = this.logLevel
@@ -190,6 +190,13 @@ class DagrCoreMain(
 
       val resources = SystemResources(cores = cores.map(Cores(_)), totalMemory = memory.map(Memory(_)))
       this.taskManager = Some(new TaskManager(taskManagerResources=resources, scriptsDirectory = scriptsDirectory, logDirectory = logDirectory))
+
+      // Print all the arguments if desired.
+      commandLine.foreach { line =>
+        config.optionallyConfigure(Configuration.Keys.PrintArgs).filter(x => x).foreach { _ =>
+          logger.info("Execution arguments: " + line)
+        }
+      }
     }
     catch {
       case v: ValidationException => throw v
--- a/sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
+++ b/sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
@@ -252,8 +252,10 @@ private[sopt] class ClpArgument(declaringClass: Class[_],
     */
   private def prettyNameValue(value: Any): String = {
     Option(value) match {
-      case Some(v) if isSensitive => String.format("--%s ***********", longName)
-      case Some(v) => String.format("--%s %s", longName, v.toString)
+      case Some(v) if isSensitive  => f"--$longName ***********"
+      case Some(Some(optionValue)) => prettyNameValue(optionValue)
+      case Some(None)              => f"--$longName ${ReflectionUtil.SpecialEmptyOrNoneToken}"
+      case Some(v)                 => f"--$longName $v"
       case _ => ""
     }
   }
@@ -265,7 +267,11 @@ private[sopt] class ClpArgument(declaringClass: Class[_],
     */
   def toCommandLineString: String = this.value match {
     case Some(v) =>
-      if (this.isCollection) prettyNameValue(new SomeCollection(v).values.mkString(" "))
+      if (this.isCollection) {
+        val someCollection = new SomeCollection(v)
+        if (someCollection.isEmpty) prettyNameValue(None)
+        else prettyNameValue(someCollection.values.mkString(" "))
+      }
       else prettyNameValue(v)
     case None =>
       throw new IllegalStateException("toCommandLineString not allowed on unset argument.")
--- a/sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
+++ b/sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
@@ -147,6 +147,12 @@ class CommandLineParser[SubCommand](val commandLineName: String)
   private type SubCommandClass = Class[_ <: SubCommand]
   private type SubCommandGroupClass = Class[_ <: ClpGroup]
 
+  private var _commandLine: Option[String] = None
+
+  /** The command line with all arguments and values stored after the [[parseSubCommand()]] or
+    * [[parseCommandAndSubCommand()]] was successful, otherwise [[None]]. */
+  def commandLine: Option[String] = _commandLine
+
   /** The error message for an unknown sub-command. */
   private[cmdline] def unknownSubCommandErrorMessage(command: String, classes: Set[SubCommandClass] = Set.empty): String = {
     s"'$command' is not a valid sub-command. See $commandLineName --help for more information." + printUnknown(command, classes.map(_.getSimpleName))
@@ -323,6 +329,7 @@ class CommandLineParser[SubCommand](val commandLineName: String)
           case ParseSuccess() => // Case 8
             val clp: SubCommand = clpParser.instance.get
             try {
+              _commandLine = Some(_commandLine.map(_ + " ").getOrElse("") + clpParser.commandLine())
               afterSubCommandBuild(clp)
               Some(clp)
             }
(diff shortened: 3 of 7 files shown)


core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -84,14 +85,15 @@ object DagrCoreMain extends Configuration {
...
-      afterSubCommandBuild            = clpBlock
+    val mainAndPipeline = parser.parseCommandAndSubCommand[DagrCoreMain](
+      args                 = args,
+      packageList          = packageList,
+      omitSubClassesOf     = Seq(classOf[DagrCoreMain]),
+      includeHidden        = includeHidden,
+      afterCommandBuild    = mainBlock,
+      afterSubCommandBuild = clpBlock
     )
+    yieldAndThen(mainAndPipeline) { mainAndPipeline.map(_._2).filter(_.printPipelineArgs).foreach { p => System.err.println(args.mkString(" ")) } }
```
tfenne: If I'm understanding this correctly, this will only print the args from the command line, and not any defaults that came from elsewhere, right?  So I think that because of where this is anything in and `@argfile` will not be present (just the literal `@argfile`) and any defaults from the class itself.

I was hoping you'd find some convenient place in `sopt` to construct a string that contained all the parameter value used to instantiate the class.

core/src/main/scala/dagr/core/tasksystem/Pipeline.scala
```diff
@@ -84,4 +84,9 @@ abstract class Pipeline(val outputDirectory: Option[Path] = None,
 
   /** True if we this pipeline is tracking this direct ancestor task, false otherwise. */
   def contains(task: Task): Boolean = tasks.contains(task)
-}
+
+  /** Override this to true to print all the command line arguments for the pipeline.  This will only print arguments
+    * if the pipeline is the top-level pipeline.
+    */
+  val printPipelineArgs: Boolean = false
```
tfenne: How about instead of this we have a global parameter in the `application.conf` that controls whether or not usages get printed, and default it to true?

core/src/test/scala/dagr/core/cmdline/DagrCoreMainTest.scala
```diff
@@ -102,6 +102,16 @@ class DagrCoreMainTest extends UnitSpec with CaptureSystemStreams with BeforeAnd
     }
   }
 
+  it should "should print the pipeline arguments if printPipelineArgs is true" in {
+    Seq(true, false).foreach { withPipelineArgsPrinter =>
+      val pipelineName = if (withPipelineArgsPrinter) nameOf(classOf[PipelineFive]) else nameOf(classOf[PipelineFour])
+      val args = Array[String](pipelineName, "--argument", "argument")
+      val (_, _, stderr, _) = testParse(args)
+      if (withPipelineArgsPrinter) stderr should include (s"$pipelineName --argument argument")
```
tfenne: Yeah, see I think we also want `--flag false` here too, so that it's recorded in case the default value changes over time.

tfenne (changes requested):
I think there's more work to do here unfortunately.  I think we want to print out a command line (or set of arguments) that includes all parameters passed to the pipeline/clp constructor, including those from argfiles and defaults, and build it in such a way that it could be used directly as a new invocation.

nh13:
@tfenne Attempt number two.  I think this changes more in the code, but use the configuration key to be set.  I omitted adding a command line argument on purpose for this.

sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
```diff
@@ -252,8 +252,10 @@ private[sopt] class ClpArgument(declaringClass: Class[_],
     */
   private def prettyNameValue(value: Any): String = {
     Option(value) match {
-      case Some(v) if isSensitive => String.format("--%s ***********", longName)
-      case Some(v) => String.format("--%s %s", longName, v.toString)
+      case Some(v) if isSensitive  => String.format("--%s ***********", longName)
```
tfenne: FYI that String.format can be replaced with `f"--${longName} ********` etc.

sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
```diff
@@ -252,8 +252,10 @@ private[sopt] class ClpArgument(declaringClass: Class[_],
...
+      case Some(Some(optionValue)) => prettyNameValue(optionValue)
+      case Some(None)              => String.format("--%s %s", longName, ReflectionUtil.SpecialEmptyOrNoneToken)
+      case Some(v)                 => String.format("--%s %s", longName, v.toString)
       case _ => ""
```
tfenne: This makes me wonder if this shouldn't return `Option[String]` and then users can `flatMap` over it to eliminate missing options, vs. this will presumably introduce extra spaces into the final CLP string or require a `filterNot(_.isEmpty)`.

sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
```diff
@@ -252,8 +252,10 @@ private[sopt] class ClpArgument(declaringClass: Class[_],
     */
   private def prettyNameValue(value: Any): String = {
     Option(value) match {
-      case Some(v) if isSensitive => String.format("--%s ***********", longName)
-      case Some(v) => String.format("--%s %s", longName, v.toString)
+      case Some(v) if isSensitive  => String.format("--%s ***********", longName)
+      case Some(Some(optionValue)) => prettyNameValue(optionValue)
+      case Some(None)              => String.format("--%s %s", longName, ReflectionUtil.SpecialEmptyOrNoneToken)
+      case Some(v)                 => String.format("--%s %s", longName, v.toString)
```
tfenne: Should we have case(s) here for collections?  I think we want `coll.mkString(" ")` as opposed to `coll.toString` which is what we'll get now.

sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
```diff
@@ -265,7 +267,11 @@ private[sopt] class ClpArgument(declaringClass: Class[_],
     */
   def toCommandLineString: String = this.value match {
```
tfenne, line 268: I see this handles the collections.  I'm not sure why it's broken into two methods though.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -417,20 +422,19 @@ class CommandLineParser[SubCommand](val commandLineName: String)
         // Get setup, and attempt to ID and load the clp class
         /////////////////////////////////////////////////////////////////////////
         val mainInstance = mainClassParser.instance.get
+        _commandLine = Some(mainClassParser.commandLine())
```
tfenne: Can I have a `this.` please?

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
```diff
@@ -147,6 +147,12 @@ class CommandLineParser[SubCommand](val commandLineName: String)
   private type SubCommandClass = Class[_ <: SubCommand]
   private type SubCommandGroupClass = Class[_ <: ClpGroup]
 
+  private var _commandLine: Option[String] = None
+
+  /** The command line with all arguments and values.  Should only be called if either [[parseSubCommand()]] or
+    * [[parseCommandAndSubCommand()]] was successful. */
+  def commandLine: String = _commandLine.get
```
tfenne: I'm curious why you don't just expose the `Option[String]` as `commandLine`?  Seems odd to store an option internally then expose a method that will throw an exception.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineProgramParser.scala
```diff
@@ -280,15 +280,16 @@ class CommandLineProgramParser[T](val targetClass: Class[T]) extends CommandLine
   }
 
   /** Gets the command line assuming `parseTasks` has been called */
-  def commandLine(printCommon: Boolean = true): String = {
+  def commandLine(): String = {
     val argumentList = this.argumentLookup.ordered.filterNot(_.hidden)
     val toolName: String = targetName
     val commandLineString: StringBuilder = new StringBuilder
 
     commandLineString.append(argumentList.filter( _.hasValue).filter(!_.isSpecial).map(_.toCommandLineString).mkString(" "))
```
tfenne: I know you didn't change it, but would you mind fixing the `filter(!_.isSpecial)` to be `filterNot(_.isSpecial)` please?

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineProgramParser.scala
```diff
@@ -280,15 +280,16 @@ class CommandLineProgramParser[T](val targetClass: Class[T]) extends CommandLine
...
     commandLineString.append(argumentList.filter(!_.hasValue).filter(!_.isSpecial).map(_.toCommandLineString).mkString(" "))
 
-    s"$toolName ${commandLineString.toString}"
+    if (commandLineString.isEmpty) toolName
+    else s"$toolName ${commandLineString.toString}"
```
tfenne: You shouldn't need `.toString` here.

sopt/src/test/scala/dagr/sopt/cmdline/ClpReflectiveBuilderTest.scala
```diff
@@ -112,4 +115,28 @@ class ClpReflectiveBuilderTest extends UnitSpec {
...
   }
+
+  it should "print v if the argument type was an Option and the value was Some(v)" in {
+    val t = new ClpReflectiveBuilder(classOf[DefaultWithSomeOption])
+    t.argumentLookup.view.foreach {
+      _.toCommandLineString shouldBe "--w 4"
+    }
+  }
+
+  it should "print the special empty or none token for a None value or empty collections" in {
```
tfenne: This doesn't sit right with me.  I would think that for `None` and empty collections we should just omit those arguments.  I could see how maybe we should print the `:none:` if the value is None/empty but the default was non-empty though.

sopt/src/test/scala/dagr/sopt/cmdline/ClpReflectiveBuilderTest.scala
```diff
@@ -112,4 +115,28 @@ class ClpReflectiveBuilderTest extends UnitSpec {
...
+    // None
+    {
+      val t = new ClpReflectiveBuilder(classOf[DefaultWithOption])
+      t.argumentLookup.view.foreach {
+        _.toCommandLineString shouldBe s"--w ${ReflectionUtil.SpecialEmptyOrNoneToken}"
+      }
+    }
+    // Empty Collection
+    {
```
tfenne: Why not just make these separate tests?  I think it would take up less space than how you've laid them out ;)

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -190,6 +190,12 @@ class DagrCoreMain(
 
       val resources = SystemResources(cores = cores.map(Cores(_)), totalMemory = memory.map(Memory(_)))
       this.taskManager = Some(new TaskManager(taskManagerResources=resources, scriptsDirectory = scriptsDirectory, logDirectory = logDirectory))
+
+      // Print all the arguments if desired.
+      commandLine match {
+        case Some(line) => config.optionallyConfigure(Configuration.Keys.PrintArgs).foreach { _ => logger.info("Execution arguments: " + line) }
```
tfenne: This isn't quite what I expected.  I was expecting to see:
```scala
config.optionallyConfigure[Boolean](Configuration.Keys.PrintArgs) match {
  Some(true) => logger.info("Execution arguments: " + line)
  _          => Unit
}
```

tfenne (approved):
Looking much better, thank you @nh13. Some minor comments and then merge when you're happy.

sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
nh13: Fixed

sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
nh13: `prettyNameValue` is private and is only used by `toCommandLineString`, with the latter handling the collection case.

sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
nh13: Deferring

sopt/src/main/scala/dagr/sopt/cmdline/ClpReflectiveBuilder.scala
nh13: We call it multiple places, and even recursively.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineParser.scala
nh13: Fixed.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineProgramParser.scala
nh13: Fixed.

sopt/src/main/scala/dagr/sopt/cmdline/CommandLineProgramParser.scala
nh13: Fixed.

sopt/src/test/scala/dagr/sopt/cmdline/ClpReflectiveBuilderTest.scala
nh13: Fixed.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
nh13: Fixed.

sopt/src/test/scala/dagr/sopt/cmdline/ClpReflectiveBuilderTest.scala
nh13: The idea is to write all the command line options, even those where the user did not override the default values. I prefer printing `:none:` when the collection is empty or the option is `None` because it is explicit.  Otherwise, when looking back on the command line, I don't know if that option defaulted to empty and so was omitted from the printed command line, or if that command line option didn't exist for the given tool/pipeline version.  I'd prefer to make it explicit that the command line option exists and its value is `:none:`.

PR #270, opened 2017-03-21
Removed setting of name in SimpleInJvmTask that replicates default behaviour.

tfenne:
@nh13 Trivial PR here.  This gets rid of the assignment in SimpleInJvmTask which clobbers the `name = getClass.getSimpleName` up in `Task`.  For `SimpleInJvmTask` itself the result is identical.  For children of `SimpleInJvmTask` they now get _their_ class name as their name instead of just `SimpleInJvmTask`.


nh13:
LGTM

PR #271, opened 2017-03-21
Fixed the messaging at the end of pipeline execution so that the warni…

tfenne:
…ng about insufficient resources is only printed if there are one or more readyTasks that cannot be scheduled.

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
@@ -553,23 +553,26 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
       allDone = graphNodesInStatesFor(List(ORPHAN, COMPLETED)).size == graphNodes.size
 
       if (!allDone && runningTasks.isEmpty && tasksToSchedule.isEmpty) {
-        logger.error(s"There are ${readyTasks.size} tasks ready to be scheduled but not enough system resources available.")
-        readyTasks.foreach { readyTask =>
-          val resourcesType: String = readyTask match {
-            case _: FixedResources    => "FixedResources"
-            case _: VariableResources => "VariableResources"
-            case _: Schedulable       => "Schedulable"
-            case _                    => "Unknown Type"
+        if (readyTasks.nonEmpty) {
+          logger.error(s"There are ${readyTasks.size} tasks ready to be scheduled but not enough system resources available.")
+          readyTasks.foreach { readyTask =>
+            val resourcesType: String = readyTask match {
+              case _: FixedResources => "FixedResources"
+              case _: VariableResources => "VariableResources"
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -553,23 +553,26 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
...
 
       if (!allDone && runningTasks.isEmpty && tasksToSchedule.isEmpty) {
-        logger.error(s"There are ${readyTasks.size} tasks ready to be scheduled but not enough system resources available.")
-        readyTasks.foreach { readyTask =>
-          val resourcesType: String = readyTask match {
-            case _: FixedResources    => "FixedResources"
-            case _: VariableResources => "VariableResources"
-            case _: Schedulable       => "Schedulable"
-            case _                    => "Unknown Type"
+        if (readyTasks.nonEmpty) {
```
nh13, line 556: I believe the only other reason to enter this branch is because there are tasks that have have unmet dependencies?  It may be useful to have a message that describes the problem if `readyTasks.isEmpty` is `true`.

nh13 (changes requested):
I know you didn't write this part, but I was wondering what message should be logged if `readTasks.isEmpty` to help the user or pipeline author along?

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
tfenne: @nh13 I'm more than happy to incorporate any messaging you want, but I'm not the author of this code (you are), and I'm honestly not sure what other cases this needs to handle.  

E.g. is it even actually possible to have tasks with unmet dependencies where the _upstream_ task is neither complete, nor "not schedulable in the available resources"?  If so, what state are they in?

nh13: I think we check for cyclical dependencies so that's not a state they are in.  Perhaps if they depend on a task that has never ben added?  I don't think an added message is critical, other then perhaps: "Unknown error" or something, just as a marker until we refactor the task system.

tfenne: Ok, well, I'm going to leave this as-is then, because the whole point of this PR is that a message was getting logged that was causing confusion by indicating that the pipeline had failed due to one reason when it was another.  Orphans are tested for above.  I think the only reason to be inside the outer `if`, and not the inner `if` here is if there are one or more `FAILED` tasks, and that isn't a scheduling error, and is reported further below anyway.  If we find some case that's not being messaged appropriately lets add it then.

nh13: +1

PR #273, opened 2017-03-22
Exposed a couple more HaplotypeCaller options.

tfenne:


tfenne:
@nh13 A trivial PR to expose a few more HC options that are useful to make it run faster (less slowly?) over very deep data.

nh13 (approved):
The mix of camelcase and underscore kill me every time. LGTM.

PR #274, opened 2017-03-25
Fixed JvmOutOfMemoryRetry so that it can find tokens that are partial…

tfenne:
… matches to lines.  Previously it requires that the token match the entire line, which never actually matched.

Also tidied up a little while I was in there.


nh13 (approved):
👍

PR #276, opened 2017-03-28
Missing space in the printed command line string.

nh13:
The command line string first has all options with values followed by all options without values. These two groups were not properly separated by a space.  Now they are!


tfenne (approved):
Looks reasonable to me.

PR #277, opened 2017-04-07
Adding a few Picard basecalling tasks.

nh13:


nh13:
@tfenne see 8e4ae73 for a travis-ci build issue fix.  Basically, the tests were using too much memory in the container environment (exit 137 from docker is a SIGKILL).

tasks/src/main/scala/dagr/tasks/picard/IlluminaBasecallsToSam.scala
```diff
@@ -0,0 +1,96 @@
...
+    Memory(this.memoryPerThread.value * minThreads) <= systemResources.systemMemory
+  }
+
+  /** Chooses the maximum # of cores given a memory per core requirement. */
+  override def pickResources(resources: ResourceSet): Option[ResourceSet] = {
+    Range.inclusive(start=maxThreads, end=minThreads, step= -1)
+      .flatMap { cores =>
+        resources.subset(Cores(cores), Memory(cores * memoryPerThread.value))
+      }.headOption
+  }
```
nh13, line 75: I am not convinced this is the right strategy.  I have seen many cases where we have A LOT more memory per core than 1G and it still wont succeed.  I am tempted to implement something that will increase the memory per core, as well as reducing the # of cores.

tasks/src/main/scala/dagr/tasks/picard/CollectIlluminaBasecallingMetrics.scala
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package dagr.tasks.picard
+
+import dagr.tasks.DagrDef.{DirPath, FilePath}
+
+import scala.collection.mutable.ListBuffer
+
+object CollectIlluminaBasecallingMetrics {
+  def toMetricsFileName(lane: Int): String = s".lane-$lane.metrics"
```
tfenne: toMetricsExtension?

tasks/src/main/scala/dagr/tasks/picard/CollectIlluminaBasecallingMetrics.scala
```diff
@@ -0,0 +1,56 @@
...
+
+class CollectIlluminaBasecallingMetrics(basecallsDir: DirPath,
+                                        lane: Int,
+                                        barcodeFile: Option[FilePath] = None,
+                                        readStructure: String,
+                                        outputDir: Option[DirPath] = None,
+                                        barcodesDir: Option[DirPath] = None
+                                       ) extends PicardTask with PicardMetricsTask {
+
+  override def in: FilePath = outputDir.getOrElse(basecallsDir).resolve("basecalling")
```
tfenne: This is bizarre.  Why would we prefer the output directory as the input directory?

tasks/src/main/scala/dagr/tasks/picard/CollectIlluminaLaneMetrics.scala
```diff
@@ -0,0 +1,52 @@
...
+import dagr.tasks.DagrDef.DirPath
+
+import scala.collection.mutable.ListBuffer
+
+/**
+  * CollectIlluminaLaneMetrics does not support read structures with molecular barcodes or skips.  If a read structure
+  * is provided with either, an error will be thrown.
+  */
+class CollectIlluminaLaneMetrics(runDirectory: DirPath,
+                                 basecallsDir: DirPath,
```
tfenne: Shouldn't this just be called `output`?

tasks/src/main/scala/dagr/tasks/picard/ExtractIlluminaBarcodes.scala
```diff
@@ -0,0 +1,76 @@
...
+                              readStructure: String,
+                              barcodeFile: FilePath,
+                              maxMismatches: Option[Int] = None,
+                              minMismatchDelta: Option[Int] = None,
+                              maxNoCalls: Option[Int] = None,
+                              minBaseQuality: Option[Int] = None,
+                              compressOutputs: Boolean = true,
+                              minThreads: Int = 4,
+                              maxThreads: Int = 16,
+                              barcodesDir: Option[DirPath] = None
```
tfenne: Again, I think this should be `output` because that's the name of the Picard parameter, and it really is just controlling where the output is going.

tasks/src/main/scala/dagr/tasks/picard/IlluminaBasecallsToSam.scala
```diff
@@ -0,0 +1,96 @@
...
+                             lane: Int,
+                             runBarcode: String,
+                             readStructure: String,
+                             libraryParamsFile: FilePath,
+                             runDate: Option[Iso8601Date] = None,
+                             sequencingCenter: Option[String] = None,
+                             includeNonPfReads: Boolean = false,
+                             minThreads: Int = 4,
+                             maxThreads: Int = 16,
+                             adapterPairs: Seq[IlluminaAdapterPair] = Seq(
```
tfenne, line 48: This seems an odd choice.  I'd either pick a more sensible default set, or leave the tool to decide the defaults.

tfenne (changes requested):
A few oddball choices perhaps informed by a single use-case, but otherwise looks ok.

tasks/src/main/scala/dagr/tasks/picard/IlluminaBasecallsToSam.scala
nh13: Solved in an upcoming commit; discussed with you.

tasks/src/main/scala/dagr/tasks/picard/CollectIlluminaBasecallingMetrics.scala
nh13: Good point, it _was_ to be able to set the output metrics file correctly, which is based on the input file name.  I could override the `PicardMetricsTask.pathPrefix` method, but instead I'll put a large comment here.

nh13: Actually, I think I can have a parameter `prefix: Option[Path]` and then it will make more sense.  I'll comment this well in the code.

tasks/src/main/scala/dagr/tasks/picard/IlluminaBasecallsToSam.scala
nh13:
```
    @Option(doc = "Which adapters to look for in the read.")
    public List<IlluminaAdapterPair> ADAPTERS_TO_CHECK = new ArrayList<IlluminaAdapterPair>(
            Arrays.asList(IlluminaAdapterPair.INDEXED,
                    IlluminaAdapterPair.DUAL_INDEXED,
                    IlluminaAdapterPair.NEXTERA_V2,
                    IlluminaAdapterPair.FLUIDIGM));
```

https://github.com/broadinstitute/picard/blob/9f41e06267010c975a5e96913ab56a101f0f2b9f/src/main/java/picard/illumina/IlluminaBasecallsToSam.java#L198-L202

nh13:
@tfenne I fixed up the `CollectIlluminaBasecallingMetrics`, improved `IlluminaBasecallsToSam` based on discussion, and fixed all but the default adapters to check (which are the ones in Picard).  Back to you.

tasks/src/main/scala/dagr/tasks/picard/CollectIlluminaLaneMetrics.scala
```diff
@@ -0,0 +1,52 @@
...
+
+package dagr.tasks.picard
+
+import dagr.tasks.DagrDef.DirPath
+
+import scala.collection.mutable.ListBuffer
+
+/**
+  * CollectIlluminaLaneMetrics does not support read structures with molecular barcodes or skips.  If a read structure
+  * is provided with either, an error will be thrown.
```
tfenne: What is the preferred solution?  To convert those operators to template?

tasks/src/main/scala/dagr/tasks/picard/ExtractIlluminaBarcodes.scala
```diff
@@ -0,0 +1,76 @@
...
+                              output: Option[DirPath] = None
+                             ) extends PicardTask with PicardMetricsTask with VariableResources {
+
+  private val _barcodesDir: DirPath = output.getOrElse(basecallsDir)
+
+  override def pickResources(resources: ResourceSet): Option[ResourceSet] = {
+    resources.subset(minCores=Cores(minThreads), maxCores=Cores(maxThreads), memory=Memory("4G"))
+  }
+
+  override def in: FilePath = _barcodesDir.resolve("barcode_counts")
```
tfenne: Could you add a similar comment here please about overriding `in` to make `metricsFile` do the right thing?  Relatedly is this actually simpler than just in `addPicardArgs` writing:
 `buffer += "METRICS_FILE=" + _barcodesDir.resolve("barcode_counts") + metricsExtension`?

tasks/src/main/scala/dagr/tasks/picard/IlluminaBasecallsToSam.scala
```diff
@@ -0,0 +1,96 @@
...
+                             barcodesDir: Option[DirPath] = None,
+                             maxReadsInRamPerTile: Option[Int] = Some(500000),
+                             tmpDir: Option[DirPath] = None
+                            ) extends PicardTask with VariableResources with Retry {
+
+  protected val byMemoryPerThread: Memory = Memory("1GB")
+  protected var memoryPerThread: Memory = Memory("2GB")
+
+  /** Increases the memory per core/thread and returns true if we can run with the fewest # of threads. */
+  override def retry(systemResources: SystemResources, taskInfo: TaskExecutionInfo): Boolean = {
```
tfenne: I wonder if you should mix in one or more of the OOM traits so that this retry only happens if an OOM was detected and not some other kind of error?

tasks/src/main/scala/dagr/tasks/picard/ExtractIlluminaBarcodes.scala
nh13: It's really to make it play nice with the `PicardMetricsTask`. I will update it to the same behavior.

nh13:
@tfenne one last look before I merge?

PR #278, opened 2017-04-08
New Exec system

nh13:
@tfenne I _think_ this is now achieves feature parity with the current execution system.  I added an option `--experimental-execution` to enable the new task system.  Interactive mode, final reports, and status logging are supported, though the latter is substantially different (and more verbose).   I have tested it on some simple workflows (see one example below).  I have a _a lot_ of cleanup to complete on the actual code, as well as likely a bunch more tests, but I am very pleased with the progress.  After cleanup, I think I am ready to add in "replay".

```scala
package dagr.pipelines

import dagr.core.cmdline.Pipelines
import dagr.core.tasksystem._
import com.fulcrumgenomics.sopt.{arg, clp}
import dagr.core.execsystem.{Cores, Memory, ResourceSet}

private trait GreedyResourcePicking extends UnitTask {
  override def pickResources(availableResources: ResourceSet): Option[ResourceSet] = {
    val mem = Memory("1g")
    (8 to 1 by -1).map(c => ResourceSet(Cores(c), mem)).find(rs => availableResources.subset(rs).isDefined)
  }
}

private class SleepProcessTask(seconds: Int = 1) extends ProcessTask with GreedyResourcePicking {
  override def args: Seq[Any] = "sleep" :: s"$seconds" :: Nil
}

private class SleepInJvmTask(seconds: Int = 1) extends SimpleInJvmTask with GreedyResourcePicking {
  def run(): Unit = Thread.sleep(seconds * 1000)
}

/**
  * Very simple example pipeline that creates random tasks and dependencies
  */
@clp(description="A bunch of sleep tasks.", group = classOf[Pipelines])
class SleepyPipeline
( @arg(flag='j', doc="Use JVM tasks") val jvmTask: Boolean = false,
  @arg(flag='n', doc="The number of tasks to create") val numTasks: Int = 100,
  @arg(flag='p', doc="The probability of creating a dependency") val dependencyProbability: Double = 0.1
) extends Pipeline {

  private def toATask: () => SleepProcessTask = () => new SleepProcessTask
  private def toBTask: () => SleepInJvmTask = () => new SleepInJvmTask
  private val toTask   = if (jvmTask) toBTask else toATask
  private val taskType = if (jvmTask) "JVM" else "Shell"

    // create the tasks
    val tasks: Seq[Task] = for (i <- 0 to numTasks) yield toTask() withName s"task-$i"

    // make them depend on previous tasks
    var rootTasks = Seq.range(start=0, numTasks).toSet
    val randomNumberGenerator = scala.util.Random
    for (i <- 0 until numTasks) {
      for (j <- 0 until i) {
        if (randomNumberGenerator.nextFloat < dependencyProbability) {
          tasks(j) ==> tasks(i)
          rootTasks = rootTasks - i
        }
      }
    }

    rootTasks.foreach { i =>
      root ==> tasks(i)
    }
  }
}
```


core/src/main/scala/dagr/core/execsystem2/Executor.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  /** An implicit to help track the failure of a future and associate it with a task status. */
+  implicit class FailWithTaskStatus[T](future: Future[T]) {
+    def failWithTaskStatus(status: TaskStatus): Future[T] = {
+      future recoverWith { case thr => Future.failed[T](TaggedException(thr=thr, status=status)) }
+    }
+  }
+}
+
+class ExecutorImpl(protected val taskExecutor: TaskExecutor) extends Executor with LazyLogging {
```
nh13: I am using `Future`s for callbacks, and I am a bit of a novice here.  But I think the idea is sound.

core/src/main/scala/dagr/core/execsystem2/Executor.scala
```diff
@@ -0,0 +1,163 @@
...
+    */
+  private def executeUnitTask(task: UnitTask): Future[Boolean] = {
+    import ExecutorImpl._
+    // Submit the task
+    this.statuses.put(task, Submitted)
+    val submitFuture: Future[TaskHandler] = this.taskExecutor.submit(task) failWithTaskStatus FailedSubmission
+
+    // Execute the task
+    val executeFuture: Future[TaskHandler] = submitFuture flatMap { handler: TaskHandler =>
+      // TODO: apply the resources it was scheduled with: task.applyResources(resources)
```
nh13: no long relevant

nh13:
@tfenne I think this is finally ready for review!

nh13:
Closing in favor of: https://github.com/fulcrumgenomics/dagr/pull/299

PR #279, opened 2017-04-09
Work in progress implementation of parsing MarkDown and converting to plain text

tfenne:
This is definitely a work in progress.   The "test" that's there is really more of a demonstration than a test, and real tests need to be written.

It does not currently support emitting terminal codes for emphasis/bold/italic, and I think to do that right would be quite a bit more work since naively inserting them into strings would screw with the word-wrapping code.  I'm not sure it's worth the effort at this point!

@nh13 Can you take a peek and see what you think of the generated output (i.e. checkout the code and run the test)?  I'll write tests and do some cleanup before I ask you to review the code.

Also, FWIW this is only using flexmark core (and flexmark util on which the core depends), which is about 500k of additional JAR.  We should probably decide if MarkDown is just what we're using, or if we want to provide MarkDown as an optional extra/plugin.

### DIFF
--- a/sopt/src/main/scala/dagr/sopt/util/MarkDownProcessor.scala
+++ b/sopt/src/main/scala/dagr/sopt/util/MarkDownProcessor.scala
@@ -1,0 +1,199 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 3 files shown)


sopt/src/main/scala/dagr/sopt/util/MarkDownProcessor.scala
```diff
@@ -0,0 +1,199 @@
...
+      *   - applying any additional gutter (e.g. to follow-on lines in list items)
+      */
+  private def indentAndWrap(chunk: Chunk): Seq[String] = {
+    val indent = " " * (indentSize * chunk.indent)
+    val gutter = indent + (" " * chunk.gutter)
+
+    if (chunk.wrappable) {
+      val length = this.lineLength - indent.length - chunk.prefix.length
+      val lines  = new ListBuffer[String]()
+      val words  = chunk.text.split("\\s+").iterator.buffered
```
tfenne, line 160: I think that if we wanted to use terminal codes for formatting the way we'd do it is introduce them in the appropriate Chunks in `process`, and then here instead of just splitting on whitespace and using the length of the text, we could split on whitespace and calculate the length of each word minus any terminal codes, and use that for wrapping calculations.

nh13:
@tfenne is this PR relevant anymore?  Closing because I don't think so.

PR #280, opened 2017-04-10
Check that the sample name does not have whitespace for VarDictJava

nh13:


tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -62,11 +62,27 @@ object VarDictJava extends Configuration {
   /** Path to the VarDictJava bin directory with a multitude of perl scripts. */
   val BinDir = RootDir.resolve("VarDict")
 
+  val IllegalCharacters: String = "\t "
```
tfenne, line 65: You could make this a regex of `\\s+` to catch all whitespace, and the...

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -62,11 +62,27 @@ object VarDictJava extends Configuration {
...
+    *
+    * @param sampleName the string that is to be used as a  sample name
+    * @param illegalCharacters the set of characters to be replaced if found, defaults to [[IllegalCharacters]]
+    * @param replacement an optional replacement character, defaulting to '_'; if [[None]] characters are just removed
+    * @return the sample name without illegal characters
+    */
+  private[vc] def replaceWhiteSpace(sampleName: String,
+                                    illegalCharacters: String = IllegalCharacters,
+                                    replacement: Option[Char] = Some('_')): String = replacement match {
+    case None    => sampleName.filter(c => !illegalCharacters.contains(c))
```
tfenne, line 83: ...down here your implementation could simply become
```scala
sampleName.replace(IllegalCharacters, replacement.getOrElse("").toString)
```

tfenne: Also, I wonder if here, or elsewhere upstream we should be `trim`ing fields, or if this problem came up in the middle of a sample name?

tfenne (commented):
I'm in two minds about this.  I don't really like that the VarDict tasks will seamlessly change the name of the sample on the fly to keep VarDict happy.  That makes things like running downstream concordance tools much more error-prone.

But then again, if VarDict can't handle sample names with spaces in them, that's poor.  What does the current failure look like?  It's not us failing to quote it on the way into VarDict right?

nh13:
@tfenne and I agreed that we would submit a PR to VarDictJava instead of this work around

tfenne:
@nh13 I have a PR in VarDict to fix this: https://github.com/AstraZeneca-NGS/VarDict/pull/47

nh13:
@tfenne I think we use `VarDictJava` (not the perl version) though.

tfenne:
We do, but VarDictJava links VarDict as a submodule, and the scripts that do the VCF creation come from the VarDict repo.  So we need to get the fix made there, and then update VarDictJava to use the latest VarDict.

nh13:
Closing in favor of https://github.com/AstraZeneca-NGS/VarDict/pull/47

PR #281, opened 2017-04-18
Add a StringUtil method addSpacesToCamelCase

nh13:
@tfenne short and sweet.

### DIFF
--- a/commons/src/main/scala/dagr/commons/util/StringUtil.scala
+++ b/commons/src/main/scala/dagr/commons/util/StringUtil.scala
@@ -24,7 +24,7 @@
 
 package dagr.commons.util
 
-import java.nio.file.{Paths, Path}
+import java.nio.file.{Path, Paths}
 import java.text.{DecimalFormat, NumberFormat}
 
 object StringUtil {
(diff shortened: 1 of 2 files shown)


commons/src/main/scala/dagr/commons/util/StringUtil.scala
```diff
@@ -119,6 +120,19 @@ object StringUtil {
     builder.toString()
   }
 
+  /** Takes in a camel-case string and places a space before any uppercase letter. */
+  def addSpacesToCamelCase(str: String): String = {
+    val builder = new StringBuilder
+    builder.sizeHint(str.length + 5)
+    val chs     = str.toCharArray
+    for (i <- 0 until chs.length) {
```
tfenne, line 128: We should really yank the `forloop` stuff down into dagr-core.

commons/src/test/scala/dagr/commons/util/StringUtilTest.scala
```diff
@@ -87,4 +87,26 @@ class StringUtilTest extends UnitSpec {
...
+  }
+
+  it should "insert spaces between adjacent capital letters" in {
+    StringUtil.addSpacesToCamelCase("TisASillyName") should be ("Tis A Silly Name")
+  }
+
+  it should "handle single letter names" in {
+    StringUtil.addSpacesToCamelCase("A") should be ("A")
+    StringUtil.addSpacesToCamelCase("b") should be ("b")
+  }
```
tfenne, line 111: Only question is do you want to add a case for multiple upper-case letters in a row?  E.g. `thisIsAThing` vs. `getSAMString`.

tfenne (approved):

commons/src/main/scala/dagr/commons/util/StringUtil.scala
nh13: Yeah.

PR #282, opened 2017-04-18
CollectBaseDistributionByCycle task

nh13:


tasks/src/main/scala/dagr/tasks/picard/CollectBaseDistributionByCycle.scala
```diff
@@ -0,0 +1,55 @@
...
+import java.nio.file.Path
+
+import dagr.tasks.DagrDef.PathToBam
+
+import scala.collection.mutable.ListBuffer
+
+object CollectBaseDistributionByCycle {
+  val MetricsExtension: String = ".base_distribution_by_cycle_metrics"
+  val ChartExtension: String = ".base_distribution_by_cycle"
+
```
tfenne: Bonus newline.

tasks/src/main/scala/dagr/tasks/picard/CollectBaseDistributionByCycle.scala
```diff
@@ -0,0 +1,55 @@
...
+package dagr.tasks.picard
+
+import java.nio.file.Path
+
+import dagr.tasks.DagrDef.PathToBam
+
+import scala.collection.mutable.ListBuffer
+
+object CollectBaseDistributionByCycle {
+  val MetricsExtension: String = ".base_distribution_by_cycle_metrics"
```
tfenne, line 35: Could we kill the `_metrics` here so we have matching file names with `.txt` and `.pdf` extensions?

tasks/src/main/scala/dagr/tasks/picard/CollectBaseDistributionByCycle.scala
```diff
@@ -0,0 +1,55 @@
...
+  val ChartExtension: String = ".base_distribution_by_cycle"
+
+}
+
+class CollectBaseDistributionByCycle(override val in: PathToBam,
+                                     override val prefix: Option[Path],
+                                     val alignedReadsOnly: Option[Boolean] = None,
+                                     val pfReadsOnly: Option[Boolean] = None)
+  extends PicardTask with PicardMetricsTask{
```
tfenne: Lacking a space before `{`

tasks/src/main/scala/dagr/tasks/picard/CollectBaseDistributionByCycle.scala
```diff
@@ -0,0 +1,55 @@
...
+  override def metricsExtension: String = ""
```
tfenne: Does this ever get invoked?  If the goal is that it shouldn't perhaps this should be `unreachable()` or something like that?

tasks/src/main/scala/dagr/tasks/picard/CollectWgsMetrics.scala
```diff
@@ -46,6 +46,6 @@ class CollectWgsMetrics(override val in: PathToBam,
     buffer.append("I=" + in)
     buffer.append("O=" + metricsFile)
     buffer.append("R=" + ref)
-    intervals.map("INTERVALS=" + _)
+    intervals.foreach(i => buffer.append("INTERVALS=" + i))
```
tfenne, line 49: Yikes, good catch.

tfenne (approved):
A bunch of trivial stuff.  Merge when you're ready.

tasks/src/main/scala/dagr/tasks/picard/CollectBaseDistributionByCycle.scala
nh13: Since most other picard metric tasks have the `_metrics` suffix, I am going to _add_ it to the chart's extension

PR #283, opened 2017-04-22
Added ability for ReflectUtil to instantiate an object using Companion.apply(String)

tfenne:
A fairly small change @nh13 once I figured it out with some help from Google.

### DIFF
--- a/commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
+++ b/commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
@@ -369,10 +369,17 @@ object ReflectionUtil {
     }
     else if (clazz == classOf[java.lang.Object]) value
     else {
-      import java.lang.reflect.Constructor
-      val ctor: Constructor[_] = clazz.getDeclaredConstructor(classOf[String])
-      ctor.setAccessible(true)
-      ctor.newInstance(value)
+      Try {
+        import java.lang.reflect.Constructor
+        val ctor: Constructor[_] = clazz.getDeclaredConstructor(classOf[String])
+        ctor.setAccessible(true)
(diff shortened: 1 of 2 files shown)


commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
```diff
@@ -369,10 +369,17 @@ object ReflectionUtil {
...
+      Try {
+        import java.lang.reflect.Constructor
+        val ctor: Constructor[_] = clazz.getDeclaredConstructor(classOf[String])
+        ctor.setAccessible(true)
+        ctor.newInstance(value)
+      }.getOrElse {
+        val typ             = mirror.classSymbol(clazz).toType
+        val companionMirror = mirror.reflectModule(typ.typeSymbol.companion.asModule)
+        val companion       = companionMirror.instance
+        companion.getClass.getMethod("apply", classOf[String]).invoke(companion, value)
```
nh13: Just a comment or two, so we remember what this does when it is not fresh in our minds.

nh13: Also, should we add documentation to `constructFromString` and other public methods that there needs to be a string constructor or `apply(str: String)` method in the companion?

nh13 (approved):
Some small suggestions on comments and documentation, otherwise 👍

commons/src/main/scala/dagr/commons/reflect/ReflectionUtil.scala
tfenne: Comment added, and I beefed up the scaladoc on both this method and the public one at the top of the calltree.

PR #286, opened 2017-04-27
Bugfix: CollectIlluminaLaneMetricsTest when converting the read struc…

nh13:
…ture.

It was not converting the read structure properly; now with tests.

I don't know why I didn't have unit tests for this previously.  Weird.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/picard/CollectIlluminaLaneMetrics.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/CollectIlluminaLaneMetrics.scala
@@ -29,6 +29,23 @@ import dagr.tasks.DagrDef.DirPath
 
 import scala.collection.mutable.ListBuffer
 
+object CollectIlluminaLaneMetrics {
+  def convertBarcodeAndSkipToTemplate(readStructure: String): String = {
+    readStructure
+      .replaceAll("[MS]", "T")
+      .split("(?<=[BT])") // split by segment type, but keep it around
+      .map(s => (s.dropRight(1).toInt, s.last.toString)) // split into template length and type
+      .foldLeft(("", 0, "")) { case ((rs, lastLen, lastSeg), (len, seg)) => // group all of the same type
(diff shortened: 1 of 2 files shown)


tasks/src/main/scala/dagr/tasks/picard/CollectIlluminaLaneMetrics.scala
```diff
@@ -29,6 +29,23 @@ import dagr.tasks.DagrDef.DirPath
 
 import scala.collection.mutable.ListBuffer
 
+object CollectIlluminaLaneMetrics {
+  def convertBarcodeAndSkipToTemplate(readStructure: String): String = {
+    readStructure
```
tfenne, line 34: I'm finding this super hard to follow - but there are tests so if you're happy I'm happy.

I think if I were to want to understand this in the future I would break it up a bit and give the intermediates really clear names and types so I could see what's going on.

tasks/src/test/scala/dagr/tasks/picard/CollectIlluminaLaneMetricsTest.scala
```diff
@@ -0,0 +1,38 @@
...
+ * THE SOFTWARE.
+ *
+ */
+
+package dagr.tasks.picard
+
+import dagr.commons.util.UnitSpec
+
+class CollectIlluminaLaneMetricsTest extends UnitSpec {
+  "CollectIlluminaLaneMetrics.convertBarcodeAndSkipToTemplate" should "convert barcodes and skips to template" in {
```
tfenne, line 31: Yay tests!

tfenne (approved):

tasks/src/main/scala/dagr/tasks/picard/CollectIlluminaLaneMetrics.scala
nh13: I find the `foldLeft` the most incomprehensible, and didn't find a way I liked to split it up.  It is what it is.

PR #289, opened 2017-05-02
Use the output prefix in CollectIlluminaBasecallingMetrics and Extrac…

nh13:
…tIlluminaBarcodes

This is so we override `prefix` in `PicardTask`.


tfenne:
👍

PR #291, opened 2017-05-05
Fixes async io usage for picard tasks; Just realized that a non-backw…

tfenne:
…ards compatible change was introduced to HTSJDK last year.


nh13 (approved):
Awesome.

PR #293, opened 2017-05-08
Removing commons and sopt from dagr and updating to 0.2.0-SNAPSHOT.

nh13:
Commons now lives here: https://github.com/fulcrumgenomics/commons
Sopt now lives here: https://github.com/fulcrumgenomics/sopt


nh13:
~~Depends on https://github.com/fulcrumgenomics/commons/pull/8~~

nh13:
@tfenne I think this PR is good-to-go.  There are a _ton_ of changes, but the most relevant is upgrading from `2.11.8` to `2.11.11`.

.travis.yml
```diff
@@ -1,7 +1,7 @@
-sudo: false
+sudo: required
 language: scala
 scala:
-- 2.11.7
+- 2.11.11
```
tfenne: Should we make this 2.12.2?  I think we want the primary build to be 2.12 with 2.11 supported by cross-build.

README.md
```diff
@@ -84,8 +84,6 @@ Tests may be run with ```sbt test```.
 You can include the five sub-projects that make up dagr using:
```
tfenne: five -> three

build.sbt
```diff
@@ -86,8 +86,8 @@ lazy val commonSettings = Seq(
   organizationHomepage := Some(url("http://www.fulcrumgenomics.com")),
   homepage             := Some(url("http://github.com/fulcrumgenomics/dagr")),
   startYear            := Some(2015),
-  scalaVersion         := "2.11.8",
-  crossScalaVersions   := Seq("2.11.8", "2.12.1"),
+  scalaVersion         := "2.11.11",
```
tfenne, line 89: Ditto here on 2.12

build.sbt
```diff
@@ -111,53 +113,24 @@ lazy val commonSettings = Seq(
...
   .settings(description := "Core methods and classes to execute tasks in dagr.")
   .settings(
     libraryDependencies ++= Seq(
-      "com.github.dblock"  %   "oshi-core"         %  "3.3",
-      "org.scala-lang"     %   "scala-reflect"     %  scalaVersion.value,
-      "org.scala-lang"     %   "scala-compiler"    %  scalaVersion.value,
-      "org.reflections"    %   "reflections"       %  "0.9.10",
-      "com.typesafe"       %   "config"            %  "1.3.1",
-      "javax.servlet"      %   "javax.servlet-api" %  "3.1.0"
+	  "com.fulcrumgenomics" %%  "commons"           %  "0.2.0-SNAPSHOT",
```
tfenne: Indent is off.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -239,8 +199,94 @@ class DagrCoreMain(
...
+      case Sopt.CommandSuccess(cmd) =>
+        unreachable("CommandSuccess should never be returned by parseCommandAndSubCommand.")
+      case Sopt.SubcommandSuccess(dagr, pipeline) =>
+        val name = pipeline.getClass.getSimpleName
+        try {
+          dagr.configure(pipeline, Some(args.mkString))
+          val name = Configuration.commandLineName(this.name)
+          printStartupLines(name, args)
+          val numFailed = dagr.execute(pipeline)
+          if (numFailed!= 0) logger.fatal("Execution failed!")
```
tfenne: Space before `!=`

tfenne: I think it's nice when something like this happens to print a larger error message.  E.g. a banner of `###`s followed by the failure, followed by more `###`s.  Just to make it more eye-catching that all did not go as planned.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -239,8 +199,94 @@ class DagrCoreMain(
...
+    logger.info(s"$name $verb. Elapsed time: $elapsedString minutes.")
+  }
+
+  /** Loads the various dagr scripts and puts them on the classpath. */
+  private def loadScripts(args: Array[String]): Unit = {
+    val tokenizer = new ArgTokenizer(args, argFilePrefix=Some("@"))
+    val collator = new ArgTokenCollator(tokenizer)
+    collator.filter(_.isSuccess).foreach {
+      case Success(ArgOptionAndValues(name: String, values: Seq[String])) if name == "scripts" =>
+        val scripts = values.map(PathUtil.pathTo(_)).filter { path =>
```
tfenne: This is a little weird to me.  Why not just filter with `Files.isReadable(_)` instead of this big try/catch?  Also, I think it would be nicer to warn or even outright fail if any provided scripts dir is unreadable.

core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -218,8 +218,8 @@ private[config] trait ConfigurationLike extends LazyLogging {
     }
   }
 
-  def commandLineName: String = {
-    configure(Configuration.Keys.CommandLineName, "Dagr")
+  def commandLineName(default: String = "dagr"): String = {
```
tfenne: It's confusing to me to have a parameter named `default` since that's a keyword in Java.  I thought for a minute this had special meaning here too.

tfenne (commented):
A few minor things here and there.  Feel free to merge when you've addressed them to your satisfaction.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
nh13: The task manager itself prints out the error messages so moving it there.  We can improve this in a different PR if the current status quo isn't what you want.

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
nh13: We can actually figure out where the the `--scripts` ends versus the pipeline name begins, so we may think that the pipeline name is a script: `java -jar dagr.core --scripts script.scala PipelineInScript`.  I suppose we could require that the scripts end with `.dagr` or `.scala`.

nh13: Also, why does `Io.assertReadable` not use `Files.isReadable` then?

PR #309, opened 2017-08-23
Added task for ClipBam in fgbio with all flag options available

maxmaliska:
Found this was needed when trying to implement ClipBam feature.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/fgbio/ClipBam.scala
+++ b/tasks/src/main/scala/dagr/tasks/fgbio/ClipBam.scala
@@ -32,24 +32,24 @@ import scala.collection.mutable.ListBuffer
 class ClipBam(val input: PathToBam,
               val output: PathToBam,
               val ref: PathToFasta,
-              val softClip: Boolean = false,
-              val autoClipAttributes: Boolean = false,
-              val readOneFivePrime: Int  = 0,
-              val readOneThreePrime: Int = 0,
-              val readTwoFivePrime: Int  = 0,
-              val readTwoThreePrime: Int = 0
+              val softClip: Option[Boolean] = None,
+              val autoClipAttributes: Option[Boolean] = None,
+              val readOneFivePrime: Option[Int] = None,
+              val readOneThreePrime: Option[Int] = None,
+              val readTwoFivePrime: Option[Int] = None,
+              val readTwoThreePrime: Option[Int] = None
              ) extends FgBioTask {
 
   override protected def addFgBioArgs(buffer: ListBuffer[Any]): Unit = {
     buffer.append("-i", input)
     buffer.append("-o", output)
     buffer.append("-r", ref)
-    buffer.append("-s", softClip)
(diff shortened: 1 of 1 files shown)


tasks/src/main/scala/dagr/tasks/fgbio/ClipBam.scala
```diff
@@ -0,0 +1,55 @@
...
+
+class ClipBam(val input: PathToBam,
+              val output: PathToBam,
+              val ref: PathToFasta,
+              val softClip: Boolean = false,
+              val autoClipAttributes: Boolean = false,
+              val readOneFivePrime: Int  = 0,
+              val readOneThreePrime: Int = 0,
+              val readTwoFivePrime: Int  = 0,
+              val readTwoThreePrime: Int = 0
```
nh13: One improvement would be to have the options `softClip` to `readTwoThreePrime` be `Option[...] = None` (ex. `val softClip: Boolean = false`) and then add to the buffer only when it is defined (ex. `softClip.foreach(s => buffer.append("-s", s))`).  This way, we don't need to put the defaults into the task (in case we change them), and so instead, the current tool defaults are used when the parameter is `None`.

nh13 (changes requested):

tasks/src/main/scala/dagr/tasks/fgbio/ClipBam.scala
maxmaliska: Sounds good. I have made that update.

tasks/src/main/scala/dagr/tasks/fgbio/ClipBam.scala
```diff
@@ -0,0 +1,55 @@
...
+              val readTwoThreePrime: Option[Int] = None
+             ) extends FgBioTask {
+
+  override protected def addFgBioArgs(buffer: ListBuffer[Any]): Unit = {
+    buffer.append("-i", input)
+    buffer.append("-o", output)
+    buffer.append("-r", ref)
+    softClip.foreach          (s => buffer.append("-s", s))
+    autoClipAttributes.foreach(a => buffer.append("-a", a))
+    readOneFivePrime.foreach  (b => buffer.append("-b", b))
```
nh13: The `readOneFivePrime` doesn't have a short flag, so would need to be specified with `--read-one-five-prime`.  The same is true for lines 50-52.

nh13 (changes requested):
Still some issues with the options

maxmaliska (commented):
I see. My mistake. I have updated it.

nh13:
A snapshot version should be built automatically with the new task shortly.  Thanks for contributing!

maxmaliska:
Great! Thanks for the quick review.

maxmaliska:
How do I know if a snapshot version has been built?

nh13:
Use `sbt update` to pull down the snapshot version, or better yet, checkout master and build it locally with `sbt publishLocal`.

PR #312, opened 2017-11-16
New task for SortBam.

tfenne:
I'm curious what your thoughts are @nh13  on how to handle the sort order param.  Dagr doesn't depend on `fgbio` currently, and I don't think we really want it to either.  So the enum isn't available.

The options as I see it are:

1. Take a String for sort order, which means that it'll always take all the values in fgbio (even if they get added to), but will also accept incorrect values
2. Make an enum here, but then if a sort order gets added in fgbio it has to get added here too

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/fgbio/SortBam.scala
+++ b/tasks/src/main/scala/dagr/tasks/fgbio/SortBam.scala
@@ -1,0 +1,48 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package dagr.tasks.fgbio
+
+import com.fulcrumgenomics.commons.io.Io
+import dagr.core.tasksystem.Pipe
+import dagr.tasks.DagrDef.PathToBam
+import dagr.tasks.DataTypes.SamOrBam
+
+import scala.collection.mutable.ListBuffer
+
+class SortBam(val in: PathToBam,
+              val out: PathToBam,
+              val sortOrder: Option[String],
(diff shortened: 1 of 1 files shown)


tasks/src/main/scala/dagr/tasks/fgbio/SortBam.scala
```diff
@@ -0,0 +1,48 @@
...
+import dagr.tasks.DagrDef.PathToBam
+import dagr.tasks.DataTypes.SamOrBam
+
+import scala.collection.mutable.ListBuffer
+
+class SortBam(val in: PathToBam,
+              val out: PathToBam,
+              val sortOrder: Option[String],
+              val maxInMemory: Option[Int]
+             ) extends FgBioTask with Pipe[SamOrBam,SamOrBam] {
```
nh13, line 38: I think it finally makes sense to add to `FgBioTask` the ability to set `--tmp-dir`, since that may matter a lot for sorting large files.  I also think we should use the `--compression` option in `FgBioTask` rather than supplying the `samjdk.compression_level` property to htsjdk.  There's also the `--async-io` option.  Sorry for making more work, but I think for this task, we'll want it.

tasks/src/main/scala/dagr/tasks/fgbio/SortBam.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  if (out == Io.StdOut) compressionLevel = Some(0)
```
nh13, line 40: What do you think about passing this into the `FgbioTask` constructor: `extends FgbioTask(compressionLevel=if (out == Io.StdOut) Some(0) else None) ...`

nh13:
I see two additional alternatives to give us access to the enum:

3. Create a new project to store type and enum definitions.  The project could be in a stand-alone repo (`bio-commons`), or we could make `fgbio` a multi-project build.  For the latter, we'd publish two artifacts, and `dagr` need only depend on the "standards/api" repo.  I assume that rarely will there be a case where we'd add to the fgbio "standards/api" without having a need for it in `fgbio`, although maybe things in `Picard` or `htsjdk` need to be exposed, though `dagr` already depends on those, so maybe not?
4. Add the enums to the `commons` project.  I don't like this idea since `commons` right now has no code that is bioinformatics specific.

I am not in favor of (2), since then we risk getting out-of-sync and the usual code-in-two-place headaches.

Currently for `fgbio` we are doing (1), but for Picard and htsjdk, we _are_ depending on them, so I would be curious to hear more about your thoughts on why we _would not_ want to depend on `fgbio` in the `dagr` `tasks` project, since "in for a penny (htsjdk/Picard), in for a pound (fgbio)".  From my perspective it is (1) or (3).

nh13 (changes requested):

tasks/src/main/scala/dagr/tasks/fgbio/SortBam.scala
tfenne: You're absolutely welcome to submit a PR to my branch to add this ;)  I think you're right, but it's totally orthogonal and not different for this task than for any other fgbio task.

tfenne:
Re: the enums and stuff.  I really dislike option (3) - that last thing I want is yet more projects and/or artifacts!  I agree with you on (4) that `commons` is really common scala code, and not anything bioinformatics-specific.  Any kind of separation like that is going to make working on fgbio more challenging, as e.g. adding a new sort would require extending the enum in one project, updating dependencies, then updating the other project.  Not to mention we sometimes tend to use enums as factories too, which wouldn't be possible.

Looking back I'm actually thinking that including Picard as a dependency was a mistake.  And given how many more dependencies htsjdk has acquired over the last year, I'm starting to feel that way there too.

Maybe the compromise position is:

1. Use `String` or `Option[String]`
2. Have non-enum constants in `dagr-tasks` so that it's easy to get the right value, but not impossible to use a value that's not enumerated?

tasks/src/main/scala/dagr/tasks/fgbio/SortBam.scala
nh13: Done.

nh13:
@tfenne lets continue with using strings and option-strings for now.  We can clean up the dependencies latter, and decide on a better approach.  It did get me thinking.

tasks/src/main/scala/dagr/tasks/fgbio/SortBam.scala
tfenne: I made this change, decided it looked more cluttered and easier to miss, then changed it back.

PR #313, opened 2017-11-16
Exposed fgbio options --async-io, --compression, and --tmp-dir

nh13:

PR #314, opened 2017-11-16
Exposed fgbio options --async-io, --compression, --tmp-dir, and --log…

nh13:
…-level.


nh13:
See #313 (oops).  Merging when build passes.

PR #315, opened 2017-11-17
add CollectHsMetrics argument for option to output PER_BASE_COVERAGE …

maxmaliska:
…file

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/picard/CollectHsMetrics.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/CollectHsMetrics.scala
@@ -34,6 +34,7 @@ import scala.collection.mutable.ListBuffer
 object CollectHsMetrics {
   val MetricsExtension   = ".hybrid_selection_metrics"
   val PerTargetExtension = ".per_target_coverage"
+  val PerBaseExtension   = ".per_base_coverage"
   def baitSetName(baitSetIntervals: Path): String = PathUtil.basename(baitSetIntervals.toString, trimExt=true)
 }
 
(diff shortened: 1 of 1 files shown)


tasks/src/main/scala/dagr/tasks/picard/CollectHsMetrics.scala
```diff
@@ -57,6 +58,7 @@ class CollectHsMetrics(override val in: PathToBam,
     buffer.append("BAIT_SET_NAME=" + baitSetName.getOrElse(CollectHsMetrics.baitSetName(targets)))
     buffer.append("LEVEL=ALL_READS")
     buffer.append("PER_TARGET_COVERAGE=" + metricsFile(extension=CollectHsMetrics.PerTargetExtension, kind=PicardOutput.Text))
+    buffer.append("PER_BASE_COVERAGE=" + metricsFile(extension=CollectHsMetrics.PerBaseExtension, kind=PicardOutput.Text))
```
nh13: I don't think we want this on by default, since it can be expensive to compute, and the output file can be large.  How about making a boolean constructor parameter to turn this on (it's off by default)?

nh13 (changes requested):

tasks/src/main/scala/dagr/tasks/picard/CollectHsMetrics.scala
maxmaliska: Sure. Makes sense. I updated with boolean that is off by default

nh13 (approved):

nh13:
@maxmaliska thanks for contributing!

PR #318, opened 2017-12-04
Updates to VarDictJava and VarDictJavaEndToEnd

nh13:
* support the -K option to include the count of Ns in the total depth: https://github.com/AstraZeneca-NGS/VarDictJava/pull/88

* support outputting all sites, including reference sites, not just sites with alternate alternate alleles.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
+++ b/tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
@@ -80,6 +80,7 @@ private class VarDictJava(tumorBam: PathToBam,
                           maximumMismatches: Option[Int] = None,
                           minimumAltReads: Option[Int] = None,
                           pileupMode: Boolean = false,
+                          countNsInTotalDepth: Boolean = false,
                           minThreads: Int = 1,
                           maxThreads: Int = 32,
                           memory: Memory = Memory("8G")
@@ -104,6 +105,7 @@ private class VarDictJava(tumorBam: PathToBam,
     minimumAltReads.foreach(buffer.append("-r", _)) // Minimum # of reads supporting an alternate allele
     minimumQuality.foreach(buffer.append("-q", _)) // The minimum base quality for a "good call"
     maximumMismatches.foreach(buffer.append("-m", _)) // Maximum number of mismatches for reads to be included, otherwise they will be filtered.
+    if (countNsInTotalDepth) buffer.append("-K") // count Ns in the total depth ("DP" field)
     buffer.append("-th", resources.cores.toInt) // The number of threads.
     buffer.append(bed)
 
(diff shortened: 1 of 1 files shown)


tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -150,20 +154,21 @@ class VarDictJavaEndToEnd
...
+      minimumQuality      = minimumQuality,
+      maximumMismatches   = maximumMismatches,
+      minimumAltReads     = minimumAltReads,
+      pileupMode          = pileupMode,
+      countNsInTotalDepth = countNsInTotalDepth,
+      minThreads          = minThreads,
+      maxThreads          = maxThreads
     )
-    val removeRefEqAltRows = new ShellCommand("awk", "{if ($6 != $7) print}") with PipeWithNoResources[Any,Any]
+    val removeRefEqAltRows = if (allSites) Pipes.empty[Any] else new ShellCommand("awk", "{if ($6 != $7) print}") with PipeWithNoResources[Any,Any]
```
tfenne, line 173: This is the only thing I'm not sure about.  This was here because VarDict is/was outputting VCF records where the same allele is listed in both the ref and alt columns, which I believe is invalid VCF and causes some implementations to fail to parse the results.  For sites that are all-ref it should output `.` in the alt column, not repeat the ref allele.  

When I did this, it was under the assumption that this was a bug.  If this is how it's outputting sites in pileup mode then I think we'll need to substitute something that, when ref == alt, sets alt to `.`.  What is VarDict outputting at reference sites?

tfenne (changes requested):
One question, hopefully not too hard to answer.

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
tfenne: From the spec: `ALT - alternate base(s): Comma separated list of alternate non-reference alleles`

nh13: @tfenne there will be a patch (not by me) to VarDict shortly to make it spec compliant in this case.  This PR is in anticipation of that.  I should have mentioned that, my apologies.

nh13: @tfenne looks like the VarDict PR has been merged!

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -150,20 +154,21 @@ class VarDictJavaEndToEnd
...
+      tumorBam            = tumorBam,
+      bed                 = bed,
+      ref                 = ref,
+      tumorName           = tn,
+      minimumAf           = minimumAf,
+      includeNonPf        = includeNonPf,
+      minimumQuality      = minimumQuality,
+      maximumMismatches   = maximumMismatches,
+      minimumAltReads     = minimumAltReads,
+      pileupMode          = pileupMode,
```
tfenne, line 168: Should we either add validation that `if (allSites) require(pileup, "...")` or have this line be `pileupMode = pileupMode || allSites`?

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
tfenne: I could find a PR in VarDict or VarDictJava that changes this behaviour, but if you're sure it's ok, go ahead.

tfenne (approved):

clintval:
I submitted this PR just now:

AstraZeneca-NGS/VarDict#58

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
nh13: Found one! https://github.com/AstraZeneca-NGS/VarDict/pull/58

nh13:
@tfenne both AstraZeneca-NGS/VarDict#58 and AstraZeneca-NGS/VarDictJava#88 have been merged, so I think this is ready to go.

PR #319, opened 2018-01-08
Clip Bam updates

nh13:


tfenne (approved):
Seems totally reasonable.

PR #326, opened 2018-06-05
DO NOT MERGE: Expose a way to modify the pipeline prior to execution.

nh13:
This may be useful to add a task to the end of all pipelines (ex.
uploading data)

### DIFF
--- a/core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
+++ b/core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
@@ -228,7 +228,8 @@ class DagrCoreMain extends LazyLogging {
         1
       case Sopt.CommandSuccess(cmd) =>
         unreachable("CommandSuccess should never be returned by parseCommandAndSubCommand.")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -228,7 +228,8 @@ class DagrCoreMain extends LazyLogging {
         1
       case Sopt.CommandSuccess(cmd) =>
         unreachable("CommandSuccess should never be returned by parseCommandAndSubCommand.")
-      case Sopt.SubcommandSuccess(dagr, pipeline) =>
+      case Sopt.SubcommandSuccess(dagr, originalPipeline) =>
+        val pipeline = modifyPipeline(dagr, originalPipeline)
```
jacarey, line 232: I think we would want to run this after configure but before execute. In the case where we want to get log and script dirs.. they will not have been set up prior to configure.

nh13: 👍 feel free to push to this branch directly

PR #328, opened 2018-06-06
Add modifyPipeline hook.

jacarey:


nh13:
LGTM, will wait for @tfenne to also bless, since this was partly my idea, so may be a bit biased.

PR #332, opened 2018-08-06
Two small-ish changes to try and make pipelines go a little faster

tfenne:
Dropped the time between scheduling runs to 250ms from 1s.
Reduced the CPU requirement for a bunch of trivial jobs to 0.1.


core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -518,17 +521,27 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
...
-      systemCores  = taskManagerResources.cores,
-      systemMemory = taskManagerResources.systemMemory,
-      jvmMemory    = taskManagerResources.jvmMemory
-    )
-    logger.debug("runSchedulerOnce: scheduling " + tasksToSchedule.size + " tasks")
-
-    // add the given tasks to the task runner with the appropriate (as determined by the scheduler) resources.
-    scheduleAndRunTasks(tasksToSchedule = tasksToSchedule)
+    val tasksToSchedule: Map[UnitTask, ResourceSet] = {
+      if (canDoAnything) {
```
nh13: It's just my own personal preference, but I prefer the `Map.empty` to come first (i.e. the small block of an `if/else`) since it is easier to read.  To find the else is many lines down in the current, but if the smaller if/else block is first then it is easier to find the `else` block.

nh13 (approved):

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
tfenne: I've inverted it for you ;)

PR #337, opened 2019-01-04
Create Var2VcfValid task and expose options in VarDictJavaEndToEnd

clintval:
I included almost all of the current flags for `var2vcf_valid.pl` in the private task class (`Var2VcfValid`) but only exposed the few options/arguments I needed for the pipeline `VarDictJavaEndToEnd`.

I just ran an integration test and can confirm that under default behavior the VCFs produced are the same before and after patch.

There are no deprecations in this PR.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
+++ b/tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
@@ -36,18 +36,17 @@ import dagr.core.tasksystem.Pipes.PipeWithNoResources
 import dagr.core.tasksystem._
 import com.fulcrumgenomics.sopt.{arg, clp}
 import dagr.tasks.DagrDef._
-import dagr.tasks.DataTypes.Vcf
 import dagr.tasks.misc.DeleteFiles
-import dagr.tasks.picard.{IntervalListToBed, SortVcf}
+import dagr.tasks.picard.SortVcf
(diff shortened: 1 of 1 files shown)


tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -171,15 +222,25 @@ class VarDictJavaEndToEnd
...
+      minimumMeanBaseQuality            = minimumMeanBaseQuality,
+      minimumDepth                      = minimumDepth,
+      minimumHighQualityAltDepth        = minimumHighQualityAltDepth,
+      minimumAf                         = minimumAf,
+      printEndTag                       = false
+    )
+
+    val sortVcf = new SortVcf(in = tmpVcf, out = out, dict = Some(dict))
+
+    root ==> (vardict | removeRefEqAltRows | bias | var2vcfvalid > tmpVcf).withName("VarDictJavaEndToEnd") ==> sortVcf ==> new DeleteFiles(tmpVcf)
```
nh13: Name "VarDictJavaEndToEnd" is also the name of the pipeline.  How about "VarDictJavaPipes" or the like?

nh13 (changes requested):
LGTM, just one comment about task naming.

clintval:
@nh13 I can't unassign me, so back to you.

Named `VarDictJavaPipeChain`.

PR #341, opened 2019-03-27
Add Picard VcfFormatConverter task

clintval:
- [x] All CLI options available in Task
- [x] I confirm works on plain `.vcf` output and compressed `.vcf.gz`/`.vcf.bgz`

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/picard/VcfFormatConverter.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/VcfFormatConverter.scala
@@ -27,11 +27,11 @@ import dagr.tasks.DagrDef.PathToVcf
 
 import scala.collection.mutable.ListBuffer
 
-class VcfFormatConverter(val in: PathToVcf, val out: PathToVcf, val requireIndex: Boolean = true) extends PicardTask {
+class VcfFormatConverter(val in: PathToVcf, val out: PathToVcf) extends PicardTask {
 
   override protected def addPicardArgs(buffer: ListBuffer[Any]): Unit = {
     buffer.append("I=" + in) // The BCF or VCF input file.
     buffer.append("O=" + out) // The BCF or VCF output file name.
-    buffer.append("REQUIRE_INDEX=" + requireIndex) // Fail if an index is not available for the input VCF/BCF.
+    buffer.append("REQUIRE_INDEX=false") // Fail if an index is not available for the input VCF/BCF.
   }
 }
--- a/tasks/src/main/scala/dagr/tasks/picard/VcfFormatConverter.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/VcfFormatConverter.scala
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2015 Fulcrum Genomics LLC
+ * Copyright (c) 2018 Fulcrum Genomics LLC
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


tasks/src/main/scala/dagr/tasks/picard/VcfFormatConverter.scala
```diff
@@ -0,0 +1,38 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 Fulcrum Genomics LLC
```
nh13: Update

tasks/src/main/scala/dagr/tasks/picard/VcfFormatConverter.scala
```diff
@@ -0,0 +1,38 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package dagr.tasks.picard
+
+import dagr.tasks.DagrDef.PathToVcf
+
+import scala.collection.mutable.ListBuffer
+
+class VcfFormatConverter(val in: PathToVcf, val out: PathToVcf, val requireIndex: Boolean = true)
```
nh13: Why is require true?

nh13 (changes requested):

tasks/src/main/scala/dagr/tasks/picard/VcfFormatConverter.scala
clintval: It's already the default of the tool but my main reason is that the developer recommends that for best results, an index file is present and `REQUIRE_INDEX` is set to `true`:
```
Converts VCF to BCF or BCF to VCF.  This tool converts files between the
plain-text VCF format and its binary compressed equivalent, BCF. Input and
output formats are determined by file extensions specified in the file names. 
For best results, it is recommended to ensure that an index file is present
and set the REQUIRE_INDEX option to true.

...

REQUIRE_INDEX=Boolean   Fail if an index is not available for the input
						VCF/BCF  Default value: true. This option can be
 						set to 'null' to clear the default value.
						Possible values: {true, false}
``` 

I chose an explicit `true` over an implicit default `true` for the above reason.

<br>

I could do the following, but since the Picard tool defaults to `true`, you can never turn off the option using a `dagr` task defined this way.

```scala
..., requireIndex: Boolean = false ...

if (requireIndex) buffer.append("REQUIRE_INDEX=true")
```

What would you prefer I do?

clintval:
Back to you, for a review of the "Variant Call Format Format Converter"

tasks/src/main/scala/dagr/tasks/picard/VcfFormatConverter.scala
clintval: I really want to avoid having to set an argument like:

```scala
doNotRequireIndex: Boolean = false
```

tfenne: So ... I'm going to go with "that documentation is bs".  `VcfFormatConverter` shouldn't even have that option - it's really very dumb.  The tool _never_ does any index querying, and doesn't touch the index other than to assert that it exists.  I say this both having a) reviewed the code and b) knowing the history of that tool due to it being someone working for me that wrote it back in the day.

I'd suggest we don't even expose the `requireIndex` parameter in dagr and simply hard code it to false in the invocation.

clintval: @tfenne, great context. Done.

tfenne:
Works for me.  @nh13 ok to merge?

PR #342, opened 2019-03-27
Adding the groupBy method to scatter/gathering

nh13:


tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -93,14 +102,21 @@ object ScatterGather {
...
       throw new UnsupportedOperationException("gather not supported on an unmapped Scatter")
 
     override def map[T <: Task](f: A => T): Scatter[T] = {
       val sub = new PrimarySubScatter(partitioner, f)
       this ==> sub
       sub
     }
+
+
+    override def groupBy[K, B <: Task](f: A => K, g: (K, Seq[A]) => B) : Scatter[B] = {
```
tfenne: I think this could probably just be another `throw new UnsupportedOperationException` because it seems very odd to me that you would scatter, do no mapping, and then immediately try and group.  If that's the case, why not just scatter on the right thing to begin with?

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -179,4 +209,46 @@ object ScatterGather {
       throw new UnsupportedOperationException("getTasks is not supported and should never be called on sub-scatters.")
     }
   }
+
+  /**
+    * A type of Scatter that is used to represent the first round of Scatter operation after the
```
tfenne: Scaladoc is copy/pasta

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -179,4 +209,46 @@ object ScatterGather {
...
+    * partitioner has created the partitions.  It's tasks are generated using the scatter partitions
+    * as input, and is also responsible for chaining together sub-scatters.
+    */
+  private class PrimaryGrouper[K, A, B <: Task](private val partitioner: Partitioner[A], f: A => K, g: (K, Seq[A]) => B) extends Pipeline with SubScatter[A,B] {
```
tfenne: This whole class can go away if you agree with the above comment about unsupporting a groupBy directly on a `PartitionerWrapper`.

tfenne (commented):
In general I like this, but I'm quibbling with myself about naming and maybe also implementation.  I understand the choice of `groupBy` given that this is in fact mapping tasks to (presumably shared) keys and then driving a new task for each group, but I tend to think that name also implies that you're _just_ grouping by, and not re-mapping/sub-scattering/sub-gathering.

Here are a few options for other names:
1. Just call it `gather` and have that method overloaded with the same signature you have for `groupBy`.  It _is_ gathering, but it's a bit weird that a `gather` would return `Scatter`.
2. Something like `subGather` or `partialGather`?
3. `reduce` actually feels fairly natural since what you're doing is reducing the cardinality of the scatter, but that also has a somewhat overloaded meaning

Another option, since `Scatter[A]` doesn't constrain the type of `A` would be to keep the name `groupBy` but simplify it to:

```scala
def groupBy[K](f: A => K): Scatter[(K,Seq[A])]
```

I think that would be a more natural definition of `GroupBy` and it could return a new subclass of `Scatter` upon which you could then `map` or `gather` etc.

Happy to chat and/or brainstorm more on this!

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -52,33 +52,42 @@ object ScatterGather {
...
      */
-    def gather[B <: Task](f: Seq[A] => B) : Gather[A,B]
+    def gather[NextResult <: Task](f: Seq[Result] => NextResult) : Gather[Result,NextResult]
+
+    /**
+      * Produces a new [[Scatter]] from the existing scatter. Takes a function that maps an individual
+      * object of type Result to a key of type K to partition the task of type Result.  Then takes a function
+      * to manufacture a [[Task]] of type NextResult from the given key of type K and partition of objects of type Result.
+      *
+      * The resulting Scatter[B] can be further mapped or gathered (or both!).
```
nh13: rephrase

clintval:
I would 💯% use this!

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -35,15 +35,15 @@ object ScatterGather {
 
   /**
    * A type of Task that, when run, partitions some kind of input, and produces one or more outputs
-   * of type A that should be parallelized (scattered) over.
+   * of type Result that should be parallelized (scattered) over.
```
nh13, line 38: A lot of the diff is changing the type parameter names FYI

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -90,84 +100,100 @@ object ScatterGather {
...
   /**
   * A type of Scatter that is used to represent the first round of Scatter operation after the
   * partitioner has created the partitions.  It's tasks are generated using the scatter partitions
   * as input, and is also responsible for chaining together sub-scatters.
   */
-  private class PrimarySubScatter[A, B <: Task](private val partitioner: Partitioner[A], f: A => B) extends Pipeline with SubScatter[A,B] {
+  private class PrimarySubScatter[Source, Result <: Task](private val partitioner: Partitioner[Source], f: Source => Result) extends Pipeline with SubScatter[Result] {
     override def build(): Unit = {
-      val as = partitioner.partitions match {
+      val as: Seq[Source] = partitioner.partitions match {
```
tfenne: Perhaps rename `as` to `sources`?

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -90,84 +100,100 @@ object ScatterGather {
...
 
   /**
   * A type of scatter that is found at stage 2 and beyond where the input is guaranteed to
   * be a Task already, and not a partition on some domain type (e.g. intervals).
   */
-  private class SecondarySubScatter[A <: Task, B <: Task](f: A => B) extends SubScatter[A,B] {
-    def chain(a: A): B = {
-      val task = f(a)
+  private class SecondarySubScatter[Source <: Task, Result <: Task](f: Source => Result) extends SubScatter[Result] {
+    def chain(a: Source): Result = {
```
tfenne: `a` -> `source`

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -179,4 +205,19 @@ object ScatterGather {
       throw new UnsupportedOperationException("getTasks is not supported and should never be called on sub-scatters.")
     }
   }
+
+  /**
+    * A type of partitioner that is found at stage 2 and beyond where the input is guaranteed to be a task already, and
+    * not part of some domain type.  The given method `f` is used to group the source tasks.
+    */
+  private class GroupByPartitioner[Result <: Task, Key](f: Result => Key) extends SimpleInJvmTask with Partitioner[(Key, Seq[Result])] {
+    var _partitions: Option[Seq[(Key, Seq[Result])]] = None
```
tfenne: Why the underscore?  Can you not just drop it and add in `override` and drop the method below?

tasks/src/test/scala/dagr/tasks/ScatterGatherTests.scala
```diff
@@ -111,4 +115,85 @@ class ScatterGatherTests extends UnitSpec with LazyLogging with BeforeAndAfterAl
...
+  private case class CountLibraries(samples: Seq[Sample], output: Path) extends SimpleInJvmTask  {
+    def run(): Unit = Io.writeLines(output, Seq(samples.map(s => (s.name, s.library)).distinct.length.toString))
+  }
+
+  private case class CountSamples(samples: Seq[Sample], output: Path) extends SimpleInJvmTask  {
+    def run(): Unit = Io.writeLines(output, Seq(samples.map(_.name).distinct.length.toString))
+  }
+
+  it should "run a scatter-gather pipeline with a group-by" in {
+    val lines = Seq("1_A 2_A 1", "1_A 2_A 2", "1_A 2_B 1", "1_B 2_A 1", "1_B 2_B 1")
```
tfenne: Could you please add a comment here on what the columns are on each line?  It would be easier than re-figuring it on re-reading.  And/or change the names to be `s1` for samples and `s1.l1` for libraries?

tfenne (approved):
Looks good.  A few minor comment, merge when you're satisfied.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13: I think I was worried that `partitions` was a public method, and I didn't want it to be set outside this scope, but I omitted making it private.  I don't think it matters since `GroupByPartitioner` is private to this scope anyhow.

PR #344, opened 2019-05-12
Add Picard BuildBamIndex to dagr tasks

clintval:
Surprised to find this little one didn't already exist!

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/picard/BuildBamIndex.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/BuildBamIndex.scala
@@ -23,11 +23,11 @@
  */
 package dagr.tasks.picard
 
-import dagr.tasks.DagrDef.PathToBam
+import dagr.tasks.DagrDef.{PathToBai, PathToBam}
 
 import scala.collection.mutable.ListBuffer
 
-class BuildBamIndex(val in: PathToBam, val out: Option[PathToBam] = None) extends PicardTask {
+class BuildBamIndex(val in: PathToBam, val out: Option[PathToBai] = None) extends PicardTask {
 
   override protected def addPicardArgs(buffer: ListBuffer[Any]): Unit = {
     buffer.append("I=" + in.toAbsolutePath)
--- a/tasks/src/main/scala/dagr/tasks/picard/BuildBamIndex.scala
+++ b/tasks/src/main/scala/dagr/tasks/picard/BuildBamIndex.scala
@@ -31,6 +31,6 @@ class BuildBamIndex(val in: PathToBam, val out: Option[PathToBai] = None) extend
 
   override protected def addPicardArgs(buffer: ListBuffer[Any]): Unit = {
     buffer.append("I=" + in.toAbsolutePath)
-    out.foreach(bai => buffer.append("O=" + bai.toString))
+    out.foreach(bai => buffer.append("O=" + bai.toAbsolutePath))
   }
 }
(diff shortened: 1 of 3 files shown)


tasks/src/main/scala/dagr/tasks/picard/BuildBamIndex.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import dagr.tasks.DagrDef.{PathToBai, PathToBam}
+
+import scala.collection.mutable.ListBuffer
+
+class BuildBamIndex(val in: PathToBam, val out: Option[PathToBai] = None) extends PicardTask {
+
+  override protected def addPicardArgs(buffer: ListBuffer[Any]): Unit = {
+    buffer.append("I=" + in.toAbsolutePath)
+    out.foreach(bai => buffer.append("O=" + bai.toString))
```
nh13: Why not?
```suggestion
    out.foreach(bai => buffer.append("O=" + bai. toAbsolutePath))
```

nh13 (changes requested):
@clintval it _**is**_ curious as to why we haven't needed this.  Most tools that produce BAM (in coordinate sorted output order), can produce them while indexing on the fly.  This removes the need to write the output BAM then read it again.  Do you have a use-case where the tool does not do that?  

Off the top of my head, if you have a tool that produces SAM, you could have it produce SAM or uncompressed BAM, then write a quick tool to which you pipe in the previous output, and the latter tool indexes and compresses it to the output.  This would make any tool that produces SAM/BAM be able to "index on the fly", as the compression/decompression is performed only once.  It also sort-of acts as asynchronous IO.  You can even have the tool sort the output data, so the SAM producing tool doesn't need to sort it.  It sound like you can just use `fgbio`'s `SortBam` or `sambamba sort`, for the latter, and some "view" tool if you don't need sorting.  `</digression>`

clintval:
@nh13 Ha I'm very much with you, but I'm working with a design decision that is not my doing.

This command will produce the following directory structure:

```bash
❯ mkdir output
❯ STAR \
	--runThreadN 6 \
	--runMode alignReads \
	--readFilesIn read1.fq read2.fq \
	--genomeDir /pipeline/references-star/hg38-100n \
	--outFileNamePrefix output/sample.library \
	--outSAMtype BAM SortedByCoordinate
```

```
❯ \ls -1 output
sample.library.Aligned.sortedByCoord.out.bam
sample.library.Log.final.out
sample.library.Log.out
sample.library.Log.progress.out
sample.library.SJ.out.tab
```

`STAR`, to the best of my knowledge, does not allow piping the alignments since `STAR` builds its own file handles internally. The best I can do is provide an output path prefix and index the BAM _after_.

Granted, I could ask `STAR` to output unsorted plaintext SAM and then perform the sort and write to binary, but that seems like a lot of writing to disk and I'd rather just skim/read the final BAM and index with a Picard tool.

clintval:
Because the BAM files `STAR` produces are also spec-non-conforming by having no read groups, I've had to find other ways of fixing the file!

https://github.com/clintval/cvbio/blob/b8fe4ba6c18b3df8e764677d2385caf5eb73c417/pipelines/src/main/scala/com/cvbio/pipeline/pipelines/StarAlignPipeline.scala#L61-L77

To be honest, I no longer truly need the `BuildBamIndex` task... but I implemented it on my way to another better method and thought it would still be a nice contribution.

I see plenty of third-party pipelines with additional `samtools index` operations and I'm not sure why exactly. I personally enjoy not having to install samtools (a second dependency) when I can do just fine with a single JAR and the base Picard routines.

PR #348, opened 2019-05-17
Optionally check configureExecutable config value path actually exists

clintval:
I have a use case which is not supported by Dagr's current methods for executable configuration.

I want to configure a default location where executables can be placed such that Dagr always uses them if they can be found. If the executables do not exist on disk at the path specified in configuration, then I want Dagr to fallback onto the system path.

With this PR, two alternate behaviors can be allowed without changing Dagr configuration. This will make sharing Dagr pipelines easier!

#### Dependency Strategies I want to Advertise

1. To get up and running very quickly, install dependencies with this command:

    ```bash
    ❯ conda install vcfanno
    ```

	Then execute the pipeline:

    ```bash
    ❯ java -jar pipeline.jar RunMe
    ```

2. To use a specific version of the pipeline dependencies or to use a different version than provided on your system path, simply place the executables at the following location:

    ```bash
    ❯ ls /pipeline/packages/
	/pipeline/packages/vcfanno 
    ```

	Then execute the pipeline:

    ```bash
    ❯ java -jar pipeline.jar RunMe
    ```

---

I'm not thrilled about providing the argument `mustExist`, would you rather this logic be allowed through a new function? Open to suggestions!

### DIFF
--- a/core/src/main/scala/dagr/core/config/Configuration.scala
+++ b/core/src/main/scala/dagr/core/config/Configuration.scala
@@ -166,24 +166,29 @@ private[config] trait ConfigurationLike extends LazyLogging {
   }
 
   /**
-    * Attempts to determine the path to an executable, first by looking it up in config,
-    * and if that fails, by attempting to locate it on the system path. If both fail then
-    * an exception is raised.
+    * Attempts to determine the path to an executable, first by looking it up in config and optionally checking to see
+    * if the path exists. If that fails, we will then locate the executable on the system path. If both strategies fail,
+    * then an exception is raised.
     *
     * @param path the configuration path to look up
     * @param executable the default name of the executable
-    * @return An absolute path to the executable to use
+    * @param mustExist ensure the configuration path to look up references a file which exists
+    * @return an absolute path to the executable to use
+    * @throws Generic when the executable cannot be configured
     */
-  def configureExecutable(path: String, executable: String) : Path = {
+  def configureExecutable(path: String, executable: String, mustExist: Boolean = false) : Path = {
     Configuration.RequestedKeys += path
 
-    optionallyConfigure[Path](path) match {
-      case Some(exec) => exec
-      case None => findInPath(executable) match {
-        case Some(exec) => exec
-        case None => throw new Generic(s"Could not configurable executable. Config path '$path' is not defined and executable '$executable' is not in PATH.")
-      }
-    }
+    optionallyConfigure[Path](path)
+      .filterNot(mustExist && !Files.exists(_))
+      .orElse(findInPath(executable))
+      .getOrElse(
+        throw new Generic(
+          s"Could not configurable executable. "
+          + s"Config path '$path' is not defined, or the config value references a file which does not exist, "
+          + s"and executable '$executable' is not in PATH."
+        )
+      )
   }
 
   /**
--- a/core/src/main/scala/dagr/core/config/Configuration.scala
+++ b/core/src/main/scala/dagr/core/config/Configuration.scala
@@ -166,9 +166,9 @@ private[config] trait ConfigurationLike extends LazyLogging {
   }
 
   /**
-    * Attempts to determine the path to an executable, first by looking it up in config and optionally checking to see
-    * if the path exists. If that fails, we will then locate the executable on the system path. If both strategies fail,
-    * then an exception is raised.
+    * Attempts to determine the path to an executable, first by looking it up in the config and optionally checking to
+    * see if the path exists. If that fails, we will then locate the executable on the system path. If both strategies
+    * fail, then an exception is raised.
     *
     * @param path the configuration path to look up
     * @param executable the default name of the executable
@@ -182,11 +182,12 @@ private[config] trait ConfigurationLike extends LazyLogging {
     optionallyConfigure[Path](path)
       .filterNot(mustExist && !Files.exists(_))
       .orElse(findInPath(executable))
+      .map(_.toAbsolutePath)
       .getOrElse(
         throw new Generic(
-          s"Could not configurable executable. "
-          + s"Config path '$path' is not defined, or the config value references a file which does not exist, "
-          + s"and executable '$executable' is not in PATH."
+          s"Could not configure executable."
+          + s" Config path '$path' is not defined, or the config value references a file which does not exist,"
+          + s" and executable '$executable' is not in the system path."
         )
       )
   }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -166,24 +166,30 @@ private[config] trait ConfigurationLike extends LazyLogging {
...
-    * and if that fails, by attempting to locate it on the system path. If both fail then
-    * an exception is raised.
+    * Attempts to determine the path to an executable, first by looking it up in the config and optionally checking to
+    * see if the path exists. If that fails, we will then locate the executable on the system path. If both strategies
+    * fail, then an exception is raised.
     *
     * @param path the configuration path to look up
     * @param executable the default name of the executable
-    * @return An absolute path to the executable to use
+    * @param mustExist ensure the configuration path to look up references a file which exists
```
nh13:
```suggestion
    * @param allowMissingConfigReference true to require that the executable path in the configuration must exist, false fallback to the system path if the configuration path exists but not the returned executable path.
```
I don't _really_ like longer names, but in this case I asked "what must exist?" prior to reading the docs, so what do think about a longer name?  Also, I tried to be more verbose about the behavior, but feel free to wordsmith further.

core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -166,24 +166,30 @@ private[config] trait ConfigurationLike extends LazyLogging {
...
+    * @return an absolute path to the executable to use
+    * @throws Generic when the executable cannot be configured
     */
-  def configureExecutable(path: String, executable: String) : Path = {
+  def configureExecutable(path: String, executable: String, mustExist: Boolean = false) : Path = {
```
nh13:
```suggestion
  def configureExecutable(path: String, executable: String, allowMissingConfigReference: Boolean = false) : Path = {
```

core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -166,24 +166,30 @@ private[config] trait ConfigurationLike extends LazyLogging {
...
 
-    optionallyConfigure[Path](path) match {
-      case Some(exec) => exec
-      case None => findInPath(executable) match {
-        case Some(exec) => exec
-        case None => throw new Generic(s"Could not configurable executable. Config path '$path' is not defined and executable '$executable' is not in PATH.")
-      }
-    }
+    optionallyConfigure[Path](path)
+      .filterNot(mustExist && !Files.exists(_))
```
nh13:
```suggestion
      .filter(!allowMissingConfigReference || Files.exists(_))
```

core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -166,24 +166,30 @@ private[config] trait ConfigurationLike extends LazyLogging {
...
+      .orElse(findInPath(executable))
+      .map(_.toAbsolutePath)
+      .getOrElse(
+        throw new Generic(
+          s"Could not configure executable."
+          + s" Config path '$path' is not defined, or the config value references a file which does not exist,"
+          + s" and executable '$executable' is not in the system path."
```
nh13: I think this statement isn't true for `allowMissingConfigReference=false`

nh13 (changes requested):
Minor changes, but I could also use @tfenne's input.  I think it may be time to move dagr out from using it's own `Configuration` class and use the [commons Configuration class](https://github.com/fulcrumgenomics/commons/blob/master/src/main/scala/com/fulcrumgenomics/commons/util/ConfigurationLike.scala).  If we do the latter, we could either add some of the missing methods directly (ex. `configureExecutable`) to the commons Configuration class, or extend it to a new trait/class (for things that want to configure executables) and put that there.

Regardless, I think this PR is fine (once comments have been addressed) and we can choose to do the migration later. @tfenne thoughts?

clintval:
I did my best wordsmithing although I feel like improvements could still be made. A re-review would be great! I agree `allowMissingConfigReference` is clearer than `mustExist` but I'm not sure `missing` is the right verb and that `ConfigReference` is the right noun. For lack of a better idea, it rests.

The parameter rename raises a few API design questions for me that might be useful context for when we eventually move to `commons.util.Configuration`.

- For a function which is destined to raise an exception when the executable can not be configured, why would we allow a state in which, under default use, a missing executable can be successfully returned?
- In the default case, do we expect the user to assert the executable exists after looking it up in the config since they will have no context as to if the executable path was found either in the config (uncertain executable existence) or on the system path (certain executable existence)?

tfenne:
I'm not entirely sure how I feel about this PR to be completely honest.  My main concern is that a mistake in config could lead to totally unexpected consequences.  E.g. I frequently have multiple versions of bwa on my system, and one of them lives on my default PATH.  If I add to config something like `bwa.executable=/packages/bwa-0.7.13/bni/bwa`, and expect pipeline will run bwa 0.7.13, I'd be unpleasantly surprised to find a) my pipelines didn't fail and b) they ran whatever version of bwa is on my system path.  That fear could be at least partly allayed by making the default behaviour to throw an error if a configured executable doesn't exist, but that means that any usages of `configureExecutable` inside of dagr wouldn't function how you'd want them.

Re: the name of `allowMissingConfigReference`, I think this is really hard to do without making it very wordy - since there are two things that could be missing; the config key could be undefined, or it's value could point to a nonexistent executable/file.  Here are a few options for (IMO) better names (with defaults that match the existing dagr behaviour):

* `configuredExecutablesMustExist: Boolean = true`
* `alwaysFallBackToSystemPath: Boolean = false`
* `allowConfiguredExecutablesToNotExist: Boolean = false`

clintval:
@tfenne, I understand your concern. Activating the feature in this PR would open you up to system path fallbacks if you included typos in your Dagr configuration.

Ultimately, I'm looking for a way to configure a Dagr pipeline using hard-coded paths as default, but permitting a system path search as a fallback if the executable is missing so I can leverage alternate dependencies that are installed using a package/environment manager like conda. This would allow me greater ease in end-to-end testing over a variety of tool dependency versions using the same Dagr pipelines (same release, same JAR).

I usually run Dagr pipelines on blank cloud instances so multiple system binaries risk is small for me but I also understand these pipelines run just as well on personal environments.

Is this PR blocked from your perspective?

Happy to keep chipping away at this feature if it's a desired addition!

nh13:
@clintval why not have an empty dagr config, then use the system path to express priority order searching for binaries?

nh13:
@tfenne @clintval  we could also implement the fallback as a global variable within `Configuration` that one could toggle.  Be default, it fails if the executable provided by the config path does not exist.  See https://github.com/fulcrumgenomics/dagr/pull/350

clintval:
Good suggestions all around. I'm 👍 on #350.

PR #350, opened 2019-05-28
Add a global configuration option to control when to fall back to system

nh13:
paths when configuring an executable.

Alternate implementation to https://github.com/fulcrumgenomics/dagr/pull/348

### DIFF
--- a/core/src/main/scala/dagr/core/config/Configuration.scala
+++ b/core/src/main/scala/dagr/core/config/Configuration.scala
@@ -46,15 +46,16 @@ object Configuration extends ConfigurationLike {
 
   // Keys for configuration values used in dagr core
   object Keys {
-    val CommandLineName = "dagr.command-line-name"
-    val ColorStatus     = "dagr.color-status"
-    val SystemPath      = "dagr.path"
-    val PackageList     = "dagr.package-list"
-    val ScriptDirectory = "dagr.script-directory"
-    val LogDirectory    = "dagr.log-directory"
-    val SystemCores     = "dagr.system-cores"
-    val SystemMemory    = "dagr.system-memory"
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -165,25 +166,40 @@ private[config] trait ConfigurationLike extends LazyLogging {
     }
   }
 
+  /** True to fall back to the searching the system path when a configuration key exists but the executable does not.
+    * False to throw an exception if the configuration key exists but the executable does not. */
+  private def fallBackToSystemPath: Boolean = configure(Configuration.Keys.FallBackToSystemPath, false)
```
nh13, line 171: I have to think about if this could be a lazy val (i.e. is config immutable)

core/src/main/scala/dagr/core/config/Configuration.scala
```diff
@@ -46,15 +46,16 @@ object Configuration extends ConfigurationLike {
...
+    val CommandLineName      = "dagr.command-line-name"
+    val ColorStatus          = "dagr.color-status"
+    val SystemPath           = "dagr.path"
+    val PackageList          = "dagr.package-list"
+    val ScriptDirectory      = "dagr.script-directory"
+    val LogDirectory         = "dagr.log-directory"
+    val SystemCores          = "dagr.system-cores"
+    val SystemMemory         = "dagr.system-memory"
+    val PrintArgs            = "dagr.print-args"
+    val FallBackToSystemPath = "dagr.fall-back-to-system-path"
```
clintval, line 58:
```
fall-back-to-system-path-for-missing-executables
```

?

PR #352, opened 2019-06-28
Adding Scatter/Gather methods flatten and flatMap, Gather method scatter, and some bug fixes

nh13:
1st Bug: https://github.com/fulcrumgenomics/dagr/pull/352/files#r298776320
2nd Bug: https://github.com/fulcrumgenomics/dagr/pull/352/files#r298776441 and https://github.com/fulcrumgenomics/dagr/pull/352/files#r298776456


core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
```diff
@@ -62,12 +62,9 @@ trait TaskTracker extends TaskManagerLike with LazyLogging {
   /** The next id for a task */
   private var nextId: TaskId = 0
 
-  /** The bi-directional map that stores the relationship between a task id and the task.  Needed to find a task
-    * given its id, since a task knows about its id, but an id does not know about its task. */
-  private val idLookup = mutable.Map[TaskId, Task]()
-
-  /** The bi-directional map that stores the relationship between a task and its execution info. */
-  private val taskAndGraphNode = new BiMap[Task, GraphNode]()
```
nh13, line 70: This is the source of the first bug.  If the input `Task` had the same `hashCode` we could get collisions (overwriting) in the `BiMap`

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -408,14 +408,17 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
...
+        node.taskInfo.submissionDate = Some(Instant.now())
         // we will make this task dependent on the tasks it creates...
         if (tasks.contains(node.task)) throw new IllegalStateException(s"Task [${node.task.name}] contained itself in the list returned by getTasks")
         // track the new tasks. If they are already added, that's fine too.
-        val taskIds: List[TaskId] = for (task <- tasks) yield addTask(task = task, enclosingNode = Some(node), ignoreExists = true)
+        val taskIds: List[TaskId] = tasks.map { task => addTask(task = task, enclosingNode = Some(node), ignoreExists = true) }
         // make this node dependent on those tasks
         taskIds.map(taskId => node.addPredecessors(this(taskId)))
+        // we may need to update precedessors if a returned task was already completed
+        if (tasks.flatMap(t => graphNodeFor(t)).exists(_.state == GraphNodeState.COMPLETED)) updatePredecessors()
```
nh13, line 421: 2nd bug (2a): if a completed task is returned by `getTasks`, the dependency graph may not be updated correctly

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -475,6 +478,8 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
           logger.exception(e, s"Failed getTasks for [${node.task.name}]: ")
       }
     }
+    // update predecessors just in case some tasks added were already completed
+    updatePredecessors()
```
nh13, line 482: 2nd bug (2b): if a completed task is returned by `getTasks`, the dependency graph may not be updated correctly

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -71,36 +81,74 @@ object ScatterGather {
...
 
+    /**
+      * Converts this [[Scatter]] of scattered scatterers into a [[Scatter]] formed by the elements of the scatterers.
+      * In other words, this scatter must be consist of elements that themselves are scatters of the resulting element
+      * type.
+      *
+      * @param toScatter an implicit conversion which asserts that the element type this scatter is a [[Scatter]].
+      * @tparam NextResult the type of element of each scattered [[Scatter]]
+      */
+    def flatten[NextResult](implicit toScatter: Result => /*<:<!!!*/ Scatter[NextResult]): Scatter[NextResult]
```
nh13: An example use case would be:
```scala
def demux(lane: Int): Scatter[Library] = ... // returns a scatter across for libraries across a given lane
def process(libraries: Seq[Library]): Sample = ...
val scatterByLane: Scatter[Int] = Scatter(partitionByLane)
val scatterByLaneDemux: Scatter[Scatter[Library]] = scatterByLane.map { lane: Int => demux(lane=lane) }
val gatherLibararies: Scatter[Library]  = scatterByLane.flatten 
// alternatively the above could be:
// gatherLibararies: Scatter[Library] =  scatterByLane.flatMap { lane: Int => demux(lane=lane)
val groupByLibrary: Scatter[(String, String), Seq[Library]) = gatherLibararies.groupBy(library => (library.sampleName, library.libraryId))
val processByLibrary: Scatter[Sample] = groupByLibrary.map { case ((sampleName, libraryId), libraries) =>
   process(libraries) 
}
```

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -71,36 +81,74 @@ object ScatterGather {
...
+      * Result to a key of type Key.  The partitions are each a sequence of objects of type Result that share the same key
+      * of type Key when the discriminator function is applied.
       *
       * The resulting Scatter[(Key, Seq[Result])] can be further mapped or gathered (or both!).
       */
     def groupBy[Key](f: Result => Key) : Scatter[(Key, Seq[Result])]
-  }
 
+    /**
+      * Converts this [[Scatter]] of scattered scatterers into a [[Scatter]] formed by the elements of the scatterers.
```
tfenne: This comment is trying far too hard to be cute.  Could just be:  "Flattens a `Scatter[Scatter[A]]` into a `Scatter[A]`."

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne: Is the `<:<!!!` in a comment meaningful here?

tfenne: I'm not entirely sure having `flatten` in the API is useful other than letting you provide a default implementation of `flatMap()`.  E.g. in your example you could just have easily redefined the following two lines:

```scala
val scatterByLaneDemux: Scatter[Scatter[Library]] = scatterByLane.map { lane: Int => demux(lane=lane) }
val gatherLibararies: Scatter[Library]  = scatterByLane.flatten 
```

as

```scala
val scatterLibrary: Scatter[Library] = scatterByLane.flatMap { lane: Int => demux(lane=lane) }
```

And the only way you could end up with a need to `flatten()` given the types involved is via `map`, so could we stick to just `flatMap`?

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -71,36 +81,74 @@ object ScatterGather {
...
   sealed trait Gather[Source, Result <: Task] extends Task {
     private[ScatterGather] var sources: Option[Seq[Source]] = None
+    private[ScatterGather] var result: Option[Result] = None
+
+    /**
+      * Produces a new [[Scatter]] from the result of the [[Gather]].
+      * @param toPartitioner the partitioner that will partition [[Result]] into one or more [[NextResult]]s.
+      * @tparam NextResult the new type of the scatter
+      */
+    def scatter[NextResult](toPartitioner: Result => Partitioner[NextResult]): Scatter[NextResult] = {
```
tfenne: So when would you use this?  When you want to re-scatter but can't define the partitions until all the upstream operations up to the gather have actually executed?  I think this needs to be made really clear in the api doc that this defers creation of the partitioner and that's the only reason you should use it over just defining a new scatter and doing `gather ==> newScatter`.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -71,36 +81,74 @@ object ScatterGather {
...
+      sub
+    }
   }
 
   /** Implementation of a Gather that performs a single All->One gather operation. */
   private class SingleGather[Source, Result <: Task](f: Seq[Source] => Result) extends Gather[Source,Result] {
     override def getTasks: Traversable[_ <: Task] = sources match {
       case None    => throw new IllegalStateException("Gather.getTasks called before sources populated.")
-      case Some(a) => Some(f(a))
+      case Some(a) => result = Some(f(a)); result
```
tfenne: I find this icky.  At a minimum it would be helpful to have `this.result = ...`.  Better would be to find a way to re-organize this so that the parent trait defines `getTasks` and then enforces that sub-classes define a method does the building and returns the result, so the `Gather` trait can catch it.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -71,36 +81,74 @@ object ScatterGather {
...
     }
   }
 
+  /** Implementation of a [[Partitioner]] that when built the upstream [[Source]] is available via the `toSource` method.
```
tfenne: I find this really hard to understand.  How about:

> A [[Partitioner]] that defers generation of the partitions until a later time, after the `Source` has become available.  Used by [[Gather.scatter]] to re-scatter dynamically after a gather operation.
> @param toSource a function that will return the Source object at a later time
> @param toPartitioner a function that will generate a partitioner once the Source is available

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
tfenne: I find all the Gather.scatter() stuff rather complicated, and I'm yet to be convinced there is a non-strawman usecase for it.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -109,6 +157,9 @@ object ScatterGather {
     override def groupBy[Key](f: Result => Key) : Scatter[(Key, Seq[Result])] =
       throw new UnsupportedOperationException("groupBy not supported on an unmapped Scatter")
 
+    override def flatten[NextResult](implicit toScatter: Result => Scatter[NextResult]): Scatter[NextResult] =
```
tfenne: Another good reason to take it out of the API and instead just have `flatMap()`

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -122,24 +173,28 @@ object ScatterGather {
     * when the partitions are available and the set of Gathers to be performed on this stage of the Scatter.
     */
   private trait SubScatter[Result <: Task] extends Scatter[Result] {
+    // the tasks that create that results
```
tfenne: No - the tasks _are_ the results.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -122,24 +173,28 @@ object ScatterGather {
...
     val tasks    = new ListBuffer[Result]
+    // the sub-scatterers that further map the results
```
tfenne: Your comment implies that there _must be_ sub-scatters (scatters, not scatterers), but that's not the case.  I.e. this is the list of _any_ sub-scatters that have been created, and may well be empty.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -122,24 +173,28 @@ object ScatterGather {
...
     val subs     = new ListBuffer[SecondarySubScatter[Result, _ <: Task]]
+    // the gatherers that gather the results
```
tfenne: Ditto, starting with "the" implies there must be gathers, but that's not the case.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -158,65 +213,88 @@ object ScatterGather {
...
+      case wrapper: PartitionerWrapper[Result] =>
+        wrapper.partitioner.partitions.getOrElse {
+          throw new IllegalStateException("Scatter/Gather pipeline trying to build before scatters populated.")
+        }
+      case _scatter => throw new IllegalStateException(s"Unknown scatter: ${_scatter.getClass.getSimpleName}")
+    }
+  }
+
+  /** A [[SecondaryPartitioner]] whose partitions are groups of sources grouped by the given method `f`. */
+  private class GroupByPartitioner[Result <: Task, Key](f: Result => Key) extends SimpleInJvmTask with SecondaryPartitioner[Result, (Key, Seq[Result])] {
```
tfenne: To me this is an oxymoron.  In the scatter/gather code:
1. A Partitioner is an operation that takes in a single thing and partitions it into many things
2. The `groupBy` operation is one that takes in many things and groups them into fewer things

Mixing the terminology (and perhaps implementation) here is incredibly confusing.

tfenne (changes requested):

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13: 1. You'll see the same comment (`<:<!!!`) in the scala collections source code, which I take to mean that `Result` must be a sub-type of `Scatter[NextResult]`!!!

2. I'll get rid of `flatten` since it should be simple going forward if we need it (when you don't start of with just a `map`).

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13: Yes, it is to partition off of the results of the gather.  I'll remove it for now :/

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13: It is gone :/

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13: That's a fact, good catch.  [Makes me think of this quote](https://en.wikipedia.org/wiki/The_medium_is_the_message)

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
nh13: I don't see anything about a `Partitioner` taking one thing, just that it produces a sequence of results (i.e. the partitions).  I think it is completely valid to take a list of things as inputs (ex. `[1,2,3,4,5]`) and produce partitions (ex. `[[1,2],[3,4],[5]]`).  That being said I've tried to rename a few things (i.e. `RePartitioner`) to make it clearer.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -38,15 +38,26 @@ object ScatterGather {
...
     def partitions: Option[Seq[Result]]
+    /** A [[Scatter]] over the resulting partitions. */
+    private[tasks] val scatter: Scatter[Result] = new PartitionerWrapper[Result](this)
+  }
+
+  /**
+    * A type of Task that, when run, re-partitions some kind of input, and produces one or more outputs
+    * of type Result that should be parallelized (scattered) over.
+    */
+  trait RePartitioner[Source <: Task, Result] extends Partitioner[Result] {
```
nh13, line 51: Trait or `sources` could be private.  What do you think @tfenne ?

nh13:
@tfenne I made almost all of the changes you suggested, with two unresolved conversations remaining.  Let's discuss offline.

tfenne (approved):
Thanks for taking the feedback!

PR #356, opened 2019-08-06
Move cross-building to 2.12 and 2.13

tfenne:
I've gotten rid of most of the deprecations except for one _huge_ class.  For some reason `Buffer.append(xs: A*)` got deprecated, and we use that literally everywhere in dagr.  To eliminate the deprecations we'd have to change e.g.:

`args.append("-i", in)`
to
`args.appendAll(Seq("-i", in))`
which just feels so much clunkier.  So I'm leaving it for now and am wondering if we should have our own arg-builder class that we use instead in some future release.

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/ResourceSet.scala
+++ b/core/src/main/scala/dagr/core/execsystem/ResourceSet.scala
@@ -59,10 +59,10 @@ case class ResourceSet(cores: Cores = Cores.none, memory: Memory = Memory.none) 
     * number of cores. Will greedily assign the highest number of cores possible.
     */
   def subset(minCores: Cores, maxCores: Cores, memory: Memory) : Option[ResourceSet] = {
-    val min = math.ceil(minCores.value).toInt
-    val max = math.floor(maxCores.value).toInt
-    val cores = max.to(min, -1).find(cores => subset(Cores(cores), memory).isDefined)
-    cores.map(c => ResourceSet(Cores(c), memory))
+    val min = minCores.value
+    val max = maxCores.value
+    val cores = Range.BigDecimal.inclusive(max, min, -1).find(cores => subset(Cores(cores.doubleValue), memory).isDefined)
+    cores.map(c => ResourceSet(Cores(c.doubleValue), memory))
   }
 
   /**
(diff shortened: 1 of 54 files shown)


pipelines/src/main/scala/dagr/pipelines/DownsampleAndCallSomaticVariants.scala
```diff
@@ -26,18 +26,17 @@ package dagr.pipelines
...
+import com.fulcrumgenomics.commons.CommonsDef._
+import com.fulcrumgenomics.commons.io.Io
 import com.fulcrumgenomics.sopt._
+import dagr.core.cmdline.Pipelines
 import dagr.core.execsystem.{Cores, Memory}
-import dagr.tasks.DagrDef
-import DagrDef._
 import dagr.core.tasksystem.{Linker, NoOpInJvmTask, Pipeline, SimpleInJvmTask}
-import com.fulcrumgenomics.commons.io.Io
+import dagr.tasks.DagrDef
```
nh13:
```suggestion
import dagr.tasks.DagrDef._
```
so you don't have the ugly `DagrDef.PathTo...` below

core/src/main/scala/dagr/core/execsystem/ResourceSet.scala
```diff
@@ -59,8 +59,8 @@ case class ResourceSet(cores: Cores = Cores.none, memory: Memory = Memory.none)
     * number of cores. Will greedily assign the highest number of cores possible.
     */
   def subset(minCores: Cores, maxCores: Cores, memory: Memory) : Option[ResourceSet] = {
-    val min = minCores.value
-    val max = maxCores.value
+    val min = math.ceil(minCores.value).toInt
+    val max = math.floor(maxCores.value).toInt
```
nh13: I am curious about this change and why it was done.  I think it is perfectly valid to have `0.9` cores available, when a task requires `0.1` cores.

nh13 (dismissed):
One potential bug introduced in `ResourceSet`.  After that, it all looks good.  Thanks for doing all this tedious work!

pipelines/src/main/scala/dagr/pipelines/DownsampleAndCallSomaticVariants.scala
tfenne: Argh.  Thanks for catching this.  I already have `CommonsDef._` imported - this was intellij being "helpful".

PR #359, opened 2019-08-13
Build a Scatter from a method that returns a sequence of results.

nh13:
Use case:

```scala
...
@arg(flag='l', doc="The lanes to analyze.") val lanes: Seq[Int]
...
{
  def build(): Unit = {
    ...
    val scatterByLanes = Scatter(this.lanes)
    ...
  }
}

```

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/ScatterGather.scala
+++ b/tasks/src/main/scala/dagr/tasks/ScatterGather.scala
@@ -30,14 +30,14 @@ import dagr.core.tasksystem.{FixedResources, Pipeline, Schedulable, SimpleInJvmT
 import scala.collection.mutable.ListBuffer
 
 /**
- * Provides a set of objects and classes for composing Scatter/Gather tasks and pipelines.
- */
+  * Provides a set of objects and classes for composing Scatter/Gather tasks and pipelines.
+  */
 object ScatterGather {
 
   /**
-   * A type of Task that, when run, partitions some kind of input, and produces one or more outputs
-   * of type Result that should be parallelized (scattered) over.
-   */
+    * A type of Task that, when run, partitions some kind of input, and produces one or more outputs
+    * of type Result that should be parallelized (scattered) over.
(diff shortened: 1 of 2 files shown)


tfenne (commented):
I like the idea in general but I'm unsure about the implementation.  A couple of things:

1. You have to be really careful using `=> Result` as a parameter type since it's _not_ a method/function reference but is more akin to a lazy val.  I can't tell just from treading the code whether the evaluation of it is deferred until the partitioner runs, or if it's evaulated sooner than that.  It would be more reliable but uglier to use `() => Result` as then you have to explicitly invoke it.

2. I wonder in general if it wouldn't be better to implement this with an alternative version of `PartitionerWrapper` that just wraps pre-made partitions?

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -56,6 +56,16 @@ object ScatterGather {
   /** Provides a factory method to create a [[Scatter]] object from a [[Partitioner]]. */
   object Scatter {
     def apply[Result](partitioner: Partitioner[Result]): Scatter[Result] = partitioner.scatter
+
+    /** Builds a [[Scatter]] from a method that returns a sequence of results. */
+    def apply[Result](toResults: => Seq[Result]): Scatter[Result] = {
```
tfenne: One last thought .. I think if you want to merge this with minimal changes, I would just change the argument type to be `Seq[Result]` and change the docs to say that it creates a scatter from a set of pre-generated partitions and call it a day.

nh13:
@tfenne I went ahead and made a task that implements `Scatter` on a fixed set of partitions.  This led to a bit of refactoring to aggregate common code and improve some of the class hierarchy.

tfenne (approved):

nh13:
Leaving this unmerged as it adds no new functionality after other PRs have been merged in.

PR #360, opened 2019-08-22
Add support for empty tasks

nh13:
1. `Tasks.getTasks` can now return an empty list of tasks to run (ex. `None`, `Seq.empty`) and the execution system will set the parent task to completed.  
2. Added a `Tasks$.empty` method to create a small task whose `getTasks` method returns `None`.

### DIFF
--- a/core/src/main/scala/dagr/core/tasksystem/Task.scala
+++ b/core/src/main/scala/dagr/core/tasksystem/Task.scala
@@ -33,6 +33,15 @@ import scala.util.control.Breaks._
 /** Utility methods to aid in working with a task. */
 object Task {
 
+  /** Marker trait for empty tasks. */
+  sealed trait EmptyTask extends Task
+
+  /** A task that does nothing. */
+  def empty: Task = new EmptyTask {
+    name = "Task.empty"
+    override def getTasks: Iterable[_ <: Task] = None
+  }
+
   /** Helper class for Tarjan's strongly connected components algorithm */
   private class TarjanData {
     var index: Int = 0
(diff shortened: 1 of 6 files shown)


core/src/main/scala/dagr/core/tasksystem/Task.scala
```diff
@@ -33,6 +33,12 @@ import scala.util.control.Breaks._
 /** Utility methods to aid in working with a task. */
 object Task {
 
+  /** A task that does nothing. */
+  def empty: Task = new Task {
```
tfenne: I think this is probably the right thing to do, but I'm also sort of curious if it would work with `val Empty = ..`, to have a single empty task.  My guess is it would probably work, but might be super confusing if the same empty task showed up at multiple very different places in a pipeline.

core/src/test/scala/dagr/core/execsystem/TaskManagerTest.scala
```diff
@@ -128,10 +128,19 @@ class TaskManagerTest extends UnitSpec with OptionValues with LazyLogging with B
     tryTaskNTimes(taskManager = taskManager, task = task, numTimes = 1, taskIsDoneFinally = true, failedAreCompletedFinally = true)
   }
 
+  it should "run an empty task" in {
```
tfenne, line 131: Might be nice to include a test that has a small pipeline of `realTask ==> Task.empty ==> anotherRealTask`.

tfenne (approved):
LGTM.

tfenne:
One other thought would be whether it would make sense to add something like `def EmptyTask: Task = Task.empty` to `Pipeline`?

core/src/main/scala/dagr/core/tasksystem/Task.scala
nh13: It would work with it being a `val`, but just in case folks (including the `Pipeline` class) want to override the name or anything else, I don't think it's a good idea to have as a `val`.

PR #362, opened 2019-08-22
Add Scatter.empty

nh13:
so we can have:
```scala
Scatter().flatMap { result => if (condition(result)) Scatter(result) else Scatter.empty
```
Also, see the added tests.

Depends on #360

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/ScatterGather.scala
+++ b/tasks/src/main/scala/dagr/tasks/ScatterGather.scala
@@ -24,7 +24,8 @@
 
 package dagr.tasks
 
-import dagr.core.tasksystem.{Pipeline, SimpleInJvmTask, Task}
+import dagr.core.execsystem.{Cores, Memory, ResourceSet}
+import dagr.core.tasksystem.{FixedResources, Pipeline, Schedulable, SimpleInJvmTask, Task, UnitTask}
 
 import scala.collection.mutable.ListBuffer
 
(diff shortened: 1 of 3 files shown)


nh13:
Added    
1.
`Scatter.apply[Result <: Task](task)` wraps a `Scatter` around a single `Task` for use with `flatMap`.  So we can do `scatter.flatMap { item => if (condition(item)) Scatter(toResult(item)) else Scatter.empty`.  Note that we could implement `Scatter.filter` as 
```scala
def filter(f: Source => Boolean): Scatter[Source] = {
  this.flatMap { source => if (f(source)) Scatter(source) else Scatter.empty }
}
```

2. 
`Scatter.apply(Iterable)` creates a Scatter from an iterable.  It was convenient for (1), but also implemented in #359

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -53,9 +54,41 @@ object ScatterGather {
     def sources: Option[Seq[Source]]
   }
 
+  /** Wraps a set of results as a [[Partitioner]] that produces those results as partitions. */
+  private class FixedPartitioner[Result](results: Iterable[Result]) extends SimpleInJvmTask with Partitioner[Result] {
+    var partitions: Option[Seq[Result]] = None
+    override def run(): Unit = this.partitions = Some(results.toSeq)
```
tfenne: I would suggest using `results.toIndexedSeq` to make sure you get a strict and not lazily evaluated seq from results.

tasks/src/main/scala/dagr/tasks/ScatterGather.scala
```diff
@@ -53,9 +54,41 @@ object ScatterGather {
...
+    val partitions: Option[Seq[Result]] = Some(Seq(result))
+    override def getTasks: Iterable[_ <: Task] = Some(result)
+  }
+
   /** Provides a factory method to create a [[Scatter]] object from a [[Partitioner]]. */
   object Scatter {
     def apply[Result](partitioner: Partitioner[Result]): Scatter[Result] = partitioner.scatter
+
+    /** Marker trait for [[Scatter.empty]]. */
+    private[tasks] sealed trait EmptyScatter[Result] extends Scatter[Result]
```
tfenne: Could you put the methods in here instead of in an anonymous subclass below.  Also, why is this a sealed trait and not just a `final class`.?

tfenne (approved):
As mentioned in slack, I think you're really redefining the scatter/gather API as a more general "working with collections of tasks" API.  Long term I think it's confusing to introduce some of these concepts while maintaining scatter/gather sematics.  If we keep pushing in this direction I think we should probably re-write this in terms of a `TaskSeq` class (which is what Scatter is becoming).

PR #365, opened 2019-08-22
Attempt to make the GATK tasks a) auto-detect GATK version and b) swi…

tfenne:
…tch command lines accordingly.

@nh13 This isn't quite done yet - firstly I haven't tested any of this and secondly I've only gone through HC and GenotypeGVCF tasks so far to adapt for 4+.  That said, I'd like feedback on the approach I've taken before I go much further.

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/gatk/GatkTask.scala
+++ b/tasks/src/main/scala/dagr/tasks/gatk/GatkTask.scala
@@ -24,6 +24,7 @@
 package dagr.tasks.gatk
 
 import java.nio.file.{Files, Path}
+import java.util.jar.Attributes.Name
 import java.util.jar.JarInputStream
 
 import dagr.core.config.Configuration
@@ -37,20 +38,23 @@ import scala.collection.mutable.ListBuffer
 object GatkTask {
   val GatkJarPathConfigKey = "gatk.jar"
 
+  private[gatk] val ImplementationVersion = new Name("Implementation-Version")
+  private[gatk] val MainClass             = new Name("Main-Class")
+
   /** Attempts to determine the major version from the GATK Jar. */
   def majorVersion(jar: FilePath): Int = {
     val in    = new JarInputStream(Files.newInputStream(jar))
     val attrs = in.getManifest.getMainAttributes
     in.close()
 
-    if (attrs.containsKey("Implementation-Version")) {
-      attrs.getValue("Implementation-Version").takeWhile(_ != '.').toInt
+    if (attrs.containsKey(ImplementationVersion)) {
+      attrs.getValue(ImplementationVersion).takeWhile(_ != '.').toInt
     }
     else {
-      attrs.getValue("Main-Class") match {
+      attrs.getValue(MainClass) match {
         case "org.broadinstitute.sting.gatk.CommandLineGATK"  => 1
         case "org.broadinstitute.gatk.engine.CommandLineGATK" => 3
-        case x => throw new IllegalArgumentException(s"Couldn't determind GATK version from jar $jar")
+        case x => throw new IllegalArgumentException(s"Couldn't determine GATK version from jar $jar")
       }
     }
   }
@@ -102,7 +106,4 @@ abstract class GatkTask(val walker: String,
 
   /** Adds arguments specific to the walker. */
   protected def addWalkerArgs(buffer: ListBuffer[Any]): Unit
-
-  /** Helper function to select an argument name for pre vs. post V4 naming. */
-  protected def either(preV4Name: String, postV4Name: String): String = if (gatkMajorVersion < 4) preV4Name else postV4Name
 }
--- a/tasks/src/main/scala/dagr/tasks/gatk/GatkTask.scala
+++ b/tasks/src/main/scala/dagr/tasks/gatk/GatkTask.scala
@@ -23,18 +23,43 @@
  */
 package dagr.tasks.gatk
 
-import java.nio.file.Path
+import java.nio.file.{Files, Path}
+import java.util.jar.JarInputStream
 
 import dagr.core.config.Configuration
 import dagr.core.execsystem.{Cores, Memory}
 import dagr.core.tasksystem.{FixedResources, ProcessTask}
-import dagr.tasks.{DagrDef, JarTask}
-import DagrDef.{PathToFasta, PathToIntervals}
+import dagr.tasks.JarTask
+import com.fulcrumgenomics.commons.CommonsDef._
 
 import scala.collection.mutable.ListBuffer
 
 object GatkTask {
   val GatkJarPathConfigKey = "gatk.jar"
+
+  /** Attempts to determine the major version from the GATK Jar. */
+  def majorVersion(jar: FilePath): Int = {
+    val in    = new JarInputStream(Files.newInputStream(jar))
+    val attrs = in.getManifest.getMainAttributes
+    in.close()
+
+    if (attrs.containsKey("Implementation-Version")) {
+      attrs.getValue("Implementation-Version").takeWhile(_ != '.').toInt
+    }
+    else {
+      attrs.getValue("Main-Class") match {
+        case "org.broadinstitute.sting.gatk.CommandLineGATK"  => 1
+        case "org.broadinstitute.gatk.engine.CommandLineGATK" => 3
+        case x => throw new IllegalArgumentException(s"Couldn't determind GATK version from jar $jar")
+      }
+    }
+  }
+
+  /** The path to the "standard" GATK jar. */
+  lazy val GatkJarPath: FilePath = Configuration.configure[Path](GatkTask.GatkJarPathConfigKey)
+
+  /** The major version of the "standard" GATK jar. */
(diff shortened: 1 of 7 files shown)


tfenne:
@clintval You might be interested in this given your recent set of tweets.  This isn't as elegant as it could be, but is an attempt to handle GATK 3.x vs. 4.x with a single set of task classes.

clintval:
I support this PR ❤️

clintval:
I spent some time thinking about how I would solve this problem in a typed workflow DSL like Dagr. I originally found inspiration in a little library that helps with [data class migrations](https://github.com/mcgizzle/oh-migrations). I dabbled with this problem in Snakemake but I [never arrived on a working solution](https://github.com/clintval/snakescale). I'm certain there is an elegant solution out there that would empower the user!

I'm thinking a first-class feature upon `ProcessTask` that would enable:

#### Backward-in-time Support

Dagr maintains a compatibility interface with older versions of the executable (similar to the aim of this PR).

#### Forward-in-time Support

Client written compatibility that provides a way to migrate an existing `ProcessTask` to a new version of the executable. Think overrides, deprecations, or translation of arguments. We would need new language features to achieve this. A concise implementation that reduces boilerplate could be tricky to conceive! If these client created migrations work well, they can be pulled into Dagr source as time marches on, bettering support for all life-time versions of an executable.

---

My free 2¢ idea

two-cents


Thanks for listening!

tasks/src/main/scala/dagr/tasks/gatk/GatkTask.scala
```diff
@@ -23,18 +23,43 @@
...
+      attrs.getValue("Implementation-Version").takeWhile(_ != '.').toInt
+    }
+    else {
+      attrs.getValue("Main-Class") match {
+        case "org.broadinstitute.sting.gatk.CommandLineGATK"  => 1
+        case "org.broadinstitute.gatk.engine.CommandLineGATK" => 3
+        case x => throw new IllegalArgumentException(s"Couldn't determind GATK version from jar $jar")
+      }
+    }
+  }
```
nh13, line 61: This might be nice to put in some form in `JarTask`.

tasks/src/main/scala/dagr/tasks/gatk/GatkTask.scala
```diff
@@ -49,18 +74,35 @@ abstract class GatkTask(val walker: String,
...
 
   /** Can be overridden to use a specific GATK jar. */
   protected def gatkJar: Path = configure[Path](GatkTask.GatkJarPathConfigKey)
 
+  /** The version of GATK being used by this task. */
+  protected lazy val gatkMajorVersion: Int = {
+    val jar = gatkJar
+    if (jar == GatkTask.GatkJarPath) GatkTask.GatkMajorVersion
+    else GatkTask.majorVersion(jar)
+  }
```
nh13, line 106: Ditto about this being something to bake into `JarTask`.  We could also have a trait too, since not everything is a JAR.

tasks/src/main/scala/dagr/tasks/gatk/GatkTask.scala
```diff
@@ -49,18 +74,35 @@ abstract class GatkTask(val walker: String,
...
+
+  /** Adds arguments specific to the walker. */
   protected def addWalkerArgs(buffer: ListBuffer[Any]): Unit
+
+  /** Helper function to select an argument name for pre vs. post V4 naming. */
+  protected def either(preV4Name: String, postV4Name: String): String = if (gatkMajorVersion < 4) preV4Name else postV4Name
```
nh13: This seems brittle if ever the GATK API changes.  And also brittle if this becomes a pattern we adopt for other tools who change their command line API more frequently.

tasks/src/main/scala/dagr/tasks/gatk/GenotypeGvcfs.scala
```diff
@@ -50,9 +50,13 @@ class GenotypeGvcfs private (ref: PathToFasta,
                              val dbSnpVcf: Option[PathToVcf] = None)
   extends GatkTask("GenotypeGVCFs", ref, intervals=intervals) {
 
+  if (gatkMajorVersion >= 4 && gvcfs.length > 1) throw new IllegalStateException(
```
nh13: `require`?

nh13 (changes requested):
I am not sure I really like this pattern.  I think there's going to be a lot of `if`/`else if`/...` logic if we follow this pattern more generally for tasks.  Let's discuss this offline.

tasks/src/main/scala/dagr/tasks/gatk/GatkTask.scala
tfenne: I think the next time we do something like this we should do that.  In an ideal world the JAR would always have `Implementation-Version` in the manifest, but it would appear that GATK versions < 4 didn't do that, so then you have to start playing tricks to ID the version.

I think also, longer term we should probably have something like `case class MultiPartVersion(major: Int, minor: Int, patch: Option[Int], wtf: Option[Int])` and use that so that you can to work with minor versions etc.  But since I could only get major for < 4, I didn't think it was worth the time to do that yet.

tasks/src/main/scala/dagr/tasks/gatk/GatkTask.scala
tfenne: Agreed ... I think I might just go back to having two separate argument generation sections for 3 vs 4 for now.

tasks/src/main/scala/dagr/tasks/gatk/HaplotypeCaller.scala
```diff
@@ -47,25 +48,47 @@ class HaplotypeCaller(ref: PathToFasta,
...
+          buffer.append("-dontUseSoftClippedBases")
+          buffer.append("-stand_call_conf", 20)
+          buffer.append("-stand_emit_conf", 20)
+        }
 
-    if (useNativePairHmm) buffer.append("-pairHMM", "VECTOR_LOGLESS_CACHING")
+      case n if n >= 4 =>
+        // Args for version 4
+        buffer.append("-I", bam.toAbsolutePath)
+        buffer.append("-O", vcf.toAbsolutePath)
```
nh13, line 79: In `GenotypeGvcfs` you have a section before this block that adds common arguments:
```scala
// Args that are common to all versions
```
I'd recommend moving common args similarly, or not have common args in `GenotypeGvcfs`

tasks/src/main/scala/dagr/tasks/gatk/GenotypeGvcfs.scala
```diff
@@ -50,9 +50,18 @@ class GenotypeGvcfs private (ref: PathToFasta,
...
+    // Args that are common to all versions
     dbSnpVcf.foreach(v => buffer.append("--dbsnp", v.toAbsolutePath.toString))
     gvcfs.foreach(gvcf => buffer.append("-V", gvcf.toAbsolutePath.toString))
-    buffer.append("-o", vcf.toAbsolutePath.toString)
+
+    gatkMajorVersion match {
+      case n if n < 4  =>
+        buffer.append("--out", vcf.toAbsolutePath.toString)
+      case n if n >= 4 =>
+        buffer.append("--output", vcf.toAbsolutePath.toString)
```
nh13: Do we need `.toString` here and elsewhere since we call it later?

tasks/src/main/scala/dagr/tasks/gatk/Mutect1.scala
```diff
@@ -51,6 +51,8 @@ class Mutect1(val tumorBam: PathToBam,
              ) extends GatkTask(walker="MuTect", ref=ref, intervals=Some(intervals)) {
 
   override protected def addWalkerArgs(buffer: ListBuffer[Any]): Unit = {
+    require(gatkMajorVersion < 4, s"Mutect v1 is not supported in GATK v${gatkMajorVersion}.")
```
nh13: This `require` is in the `addWalkerArgs` method, but in `IndelRealigner` and `RealignerTargetCreator` it's in the constructor.  The former will check when we run the task, while the latter when we build the task.  Up to you.

nh13 (approved):
Minor quibbles, fix/resolve and approved.

tasks/src/main/scala/dagr/tasks/gatk/GenotypeGvcfs.scala
tfenne: I have no idea why there were `.toString`s there in the first place.  Removed.

tasks/src/main/scala/dagr/tasks/gatk/HaplotypeCaller.scala
tfenne: I almost did it, but in fact there's only _1_ common arg, and it felt not worth it.  Note that output is `-o` vs.`-O`.  But I can move the `-I` argument out.

tasks/src/main/scala/dagr/tasks/gatk/Mutect1.scala
tfenne: Good call.  I'll move this one up because failing earlier seems better.

PR #367, opened 2019-10-02
PicardTask to allow overriding default use of JDK inflater

clintval:


nh13 (approved):
This looks good to me, any reason why you didn't also add `USE_JDK_DEFLATER` while you were there?  For other folks, this is related to the following issues:
* https://github.com/samtools/htsjdk/issues/1420
* https://github.com/broadinstitute/picard/issues/1383
* https://github.com/Intel-HLS/GKL/issues/101

clintval:
I think that's what I did? If the Boolean is `true`, the JDK inflater is used, if the Boolean is `false`, the Intel inflater (if available) is used. If no Boolean is provided, the Picard default is used.

PR #368, opened 2019-10-08
PicardTask to use the Intel inflater/deflater only if it is safe, by default

clintval:
Many of my Dagr pipelines now require the Intel Inflater/Deflater to be conditionally turned off across all executables that use htsjdk because of Intel-HLS/GKL#101. Can we intelligently use the Intel Inflater/Deflater, or not, on any Picard task as a default?

Followup to: https://github.com/fulcrumgenomics/dagr/pull/367 (including your comment) and https://github.com/fulcrumgenomics/fgbio/pull/534

**Note**: None of my recent PRs provide parameterization, or default-setting, for GATK tasks!

I have tested this using pipelines that have Picard tasks and confirm it works as expected!

### DIFF
--- a/core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
+++ b/core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
@@ -43,48 +43,6 @@ import dagr.core.tasksystem.Pipeline
 import scala.collection.mutable.ListBuffer
 import scala.util.{Failure, Success}
 
-
-/** A little class so that we don't have to rely on SystemUtils from org.apache.commons:commons-lang3.
-  *
-  * This implementation is duplicated from `com.fulcrumgenomics.cmdline.SystemUtils`.
-  */
-private[dagr] object SystemUtils {
-  /** The OS name prefixes for Linux */
-  private val LinuxNamePrefixes: Seq[String] = Seq("Linux", "LINUX")
-  /** The OS name prefixes for Mac */
-  private val MacNamePrefixes: Seq[String]   = Seq("Mac")
-  /** The current OS name. */
-  private val OsName: Option[String]         = getSystemProperty("os.name")
-  /** The current OS architecture */
-  private val OsArch: Option[String]         = getSystemProperty("os.arch")
-  /** The current OS version */
-  private val OsVersion: Option[String]      = getSystemProperty("os.version")
-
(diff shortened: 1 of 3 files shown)


core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
```diff
@@ -43,6 +43,48 @@ import dagr.core.tasksystem.Pipeline
...
+  private def IsOsArch(name: String): Boolean = OsArch.contains(name)
+  /** Returns true if the operating system version starts with the given version string, false otherwise. */
+  private def IsOsVersion(prefix: String): Boolean = OsVersion.exists(_.startsWith(prefix))
+
+  /** True if the current system supports the Intel Inflater and Deflater, false otherwise. */
+  val IntelCompressionLibrarySupported: Boolean = {
+    if (!SystemUtils.IsOsLinux && !SystemUtils.IsOsMac) false
+    else if (SystemUtils.IsOsArch("ppc64le")) false
+    else if (SystemUtils.IsOsMac && SystemUtils.IsOsVersion("10.14.")) false // FIXME: https://github.com/Intel-HLS/GKL/issues/101
+    else true
```
nh13: If you’ve duplicated code, it goes in commons. That being said, why don’t you patch Picard?

nh13 (changes requested):

core/src/main/scala/dagr/core/cmdline/DagrCoreMain.scala
clintval: I was hesitant to place in commons since commons is a place for "Common scala utilities". I can surely do so when I find the time.

Patching Picard is not exclusive with solving this at the workflow level. Many of my Picard JARs are not bleeding edge.

clintval: PR here: https://github.com/fulcrumgenomics/commons/pull/59

nh13:
@clintval this should be unblocked given fulcrumgenomics/commons#59 is merged.

clintval:
@nh13 all set. Is TravisCI no longer enabled in this repository?

nh13:
Re-opening to see if I can get travis-ci to build.

nh13:
@clintval looks like it wasn't triggering, re-opening did the trick

PR #372, opened 2020-02-10
Add the "-K" option from bwa

nh13:


tasks/src/main/scala/dagr/tasks/bwa/Bwa.scala
```diff
@@ -49,15 +49,16 @@ object Bwa extends Configuration {
                      mergeBamAlignmentMem: String = "2G",
                      fifoBufferMem: String = "512M",
                      processAltMappings: Boolean = false,
-                     orientation: PairOrientation = PairOrientation.FR
+                     orientation: PairOrientation = PairOrientation.FR,
+                     numFixedBases: Option[Int] = None
```
tfenne: Maybe `numInputBases`?  Or `basesPerBatch`?

tfenne (approved):
Approved but with some suggestions for names.

PR #374, opened 2020-03-23
ResourceSet bug when subsetting resources and some performance optimizations

nh13:


core/src/main/scala/dagr/core/execsystem/ResourceSet.scala
```diff
@@ -56,13 +56,30 @@ case class ResourceSet(cores: Cores = Cores.none, memory: Memory = Memory.none)
...
+    * Example 2: minCores=1, maxCores=5.1, this.cores=4.5, then 4.5 cores will be returned.
+    * Example 3: minCores=1, maxCores=5, this.cores=1.5, then 1 core will be returned.
+    * Example 4: minCores=1.5, maxCores=5, this.cores=1.5, then 1.5 cores will be returned.
     */
   def subset(minCores: Cores, maxCores: Cores, memory: Memory) : Option[ResourceSet] = {
-    val min = minCores.value
-    val max = maxCores.value
-    val cores = Range.BigDecimal.inclusive(max, min, -1).find(cores => subset(Cores(cores.doubleValue), memory).isDefined)
-    cores.map(c => ResourceSet(Cores(c.doubleValue), memory))
+    if (!subsettable(ResourceSet(minCores, memory))) None else {
```
nh13, line 70: part of me feels like I am missing something.  Why was `Range.BigDecimal.inclusive` ever used when we can know directly what to use?

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -559,8 +563,8 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
     val runningTasks: Map[UnitTask, ResourceSet] = runningTasksMap
 
     // get the tasks that are eligible for execution (tasks with no dependents)
-    val (emptyTasks: List[Task], readyTasks: List[Task]) = {
-      graphNodesInStateFor(NO_PREDECESSORS).map(_.task).toList.partition(_.isInstanceOf[Task.EmptyTask])
+    val (emptyTasks: Seq[Task], readyTasks: Seq[Task]) = {
+      graphNodesInStateFor(NO_PREDECESSORS).map(_.task).toSeq.partition(_.isInstanceOf[Task.EmptyTask])
```
nh13, line 567: As a side note: I really think we can speed up dagr further by maintaining a bit more internal state.  `graphNodesInStateFor` loops through all the nodes, versus just having a simple lookup.  This code really shows how poor my first attempt at Scala was.

core/src/main/scala/dagr/core/execsystem/ResourceSet.scala
tfenne: I think it was just a tidy way to express what was going on, and likely evolved from when Cores was an `Int`.  I don't think either of us expected anyone to pass in values large enough (>> 1000) for it to cause a problem.

core/src/main/scala/dagr/core/execsystem/ResourceSet.scala
```diff
@@ -56,13 +56,30 @@ case class ResourceSet(cores: Cores = Cores.none, memory: Memory = Memory.none)
...
+      val coresValue = {
+        if (maxCores.value == math.floor(maxCores.value)) { // return a whole number
```
tfenne: I think `maxCores.value.isValidInt` is likely clearer here.

tfenne: I'm wondering if the right answer here is to make this explicit instead?  E.g. `subset(minCores, maxCores, allowPartialCores: Boolean = false, memory)`?  Because while the old implementation may not have supported it it would be nice to be able to specify `subset(minCores=1.5, maxCores=100, memory="6G")` and have the job be scheduled even if there are only 1.5 cores available.

The other option would be to check for both `minCores` and `maxCores` and allow partial cores if _either_ is non-integer?

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
tfenne: Yikes.  Maybe log an issue to come back and fix this some day?

core/src/main/scala/dagr/core/execsystem/TaskTracker.scala
```diff
@@ -315,4 +315,9 @@ trait TaskTracker extends TaskManagerLike with LazyLogging {
   protected def graphNodesWithPredecessors: Iterable[GraphNode] = {
     graphNodes.filter(node => GraphNodeState.hasPredecessors(node.state))
   }
+
+  protected def forallGraphNodes(states: Set[GraphNodeState.Value]): Boolean = {
```
tfenne: Since this takes a set of states rather than a predicate, how about `allGraphNodesInStates()`?

tfenne (approved):

core/src/main/scala/dagr/core/execsystem/ResourceSet.scala
nh13: @tfenne the last suggestion is exactly what I am trying to do.  Perhaps the docs aren't clear?

PR #376, opened 2020-04-03
Add withPrefix and withSuffix to Pipeline

mjhipp:
I am suggesting adding the ability to change or set the prefix or suffix of a Pipeline at or after construction time. I believe it can help to make logs more readable. Modeled after `Task.withName`.


nh13:
thanks @mjhipp !

PR #378, opened 2020-04-27
Add VarDictJava's experimental --fisher option to VarDictJavaEndToEnd

mjhipp:
VarDictJava added a new feature to move some of the post-processing from R to Java in https://github.com/AstraZeneca-NGS/VarDictJava/pull/278. In that PR, they state that the R scripts will increase exponentially with larger VCF files, while the Java calculations do not. For a 3MB BAM they see no speedup, but for a 17GB BAM they save a couple hours, and for a 30GB BAM, they save a couple **days**. 

This PR adds that option to the `VarDictJavaEndToEnd` pipeline. 

I ran `VarDictJavaEndToEnd` on a couple smaller BAMs and these were the runtimes:

| BAM size | Runtime using R Fisher | Runtime using Java Fisher | 
|:-:|:--:|:--:|
| ~13MB | 11 seconds | 11 seconds |
| ~400MB | 3.18 minutes | 2.61 minutes |

The only differences in the resulting VCF files were that the Java Fisher output uses decimal notation and the R version uses scientific notation.

```diff
< AF=3e-03
> AF=0.003 
```

If the new option is used on an older version of `VarDictJava`, we get the following error (which is expected):

```console
Exception in thread "main" org.apache.commons.cli.UnrecognizedOptionException: Unrecognized option: --fisher
	at org.apache.commons.cli.Parser.processOption(Parser.java:363)
	at org.apache.commons.cli.Parser.parse(Parser.java:199)
	at org.apache.commons.cli.Parser.parse(Parser.java:85)
	at com.astrazeneca.vardict.CmdParser.parseParams(CmdParser.java:31)
	at com.astrazeneca.vardict.Main.main(Main.java:13)

```

### DIFF
--- a/tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
+++ b/tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
@@ -125,6 +125,7 @@ private class VarDictJava(tumorBam: PathToBam,
                           minimumAltReads: Option[Int] = None,
                           pileupMode: Boolean = false,
                           countNsInTotalDepth: Boolean = false,
+                          fisher: Boolean = false,
                           minThreads: Int = VarDictJava.MinimumThreads,
                           maxThreads: Int = VarDictJava.MaximumThreads,
                           memory: Memory = VarDictJava.DefaultMemory
@@ -163,6 +164,7 @@ private class VarDictJava(tumorBam: PathToBam,
     minimumQuality.foreach(buffer.append("-q", _)) // The minimum base quality for a "good call".
     maximumMismatches.foreach(buffer.append("-m", _)) // Maximum number of mismatches for reads to be included, otherwise they will be filtered.
     if (countNsInTotalDepth) buffer.append("-K") // Count Ns in the total depth ("DP" field).
+    if (fisher) buffer.append("--fisher") // Changes R script (teststrandbias.R and testsomatic.R) to Java implementation of Fisher exact test. Requires VarDictJava version 1.8.0 (b772179) or later.
     buffer.append("-th", resources.cores.toInt) // The number of threads.
     buffer.append(bed)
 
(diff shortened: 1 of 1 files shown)


tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -314,7 +316,9 @@ class VarDictJavaEndToEnd
  @arg(flag='T', doc="The maximum # of threads with which to run.") maxThreads: Int = VarDictJava.MaximumThreads,
  @arg(flag='a', doc="Output all sites, including reference calls.") allSites: Boolean = false,
  @arg(flag='A', doc="Output all variants at the same genomic site.") allVariants: Boolean = false,
- @arg(flag='N', doc="Count No-calls (Ns) in the total depth tag (DP)") countNsInTotalDepth: Boolean = false) extends Pipeline {
+ @arg(flag='N', doc="Count No-calls (Ns) in the total depth tag (DP)") countNsInTotalDepth: Boolean = false,
+ @arg(flag='F', doc="Experimental feature: Changes R script (teststrandbias.R and testsomatic.R) to Java implementation of Fisher exact test. Requires VarDictJava version 1.8.0 (b772179) or later.") fisherJava: Boolean = false
```
nh13:
```suggestion
 @arg(flag='F', doc="Experimental feature: Use Java implementation of Fisher exact test (previously R scripts:teststrandbias.R and testsomatic.R). Requires VarDictJava version 1.8.0 (b772179) or later.") fisherJava: Boolean = false
```

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -125,6 +125,7 @@ private class VarDictJava(tumorBam: PathToBam,
                           minimumAltReads: Option[Int] = None,
                           pileupMode: Boolean = false,
                           countNsInTotalDepth: Boolean = false,
+                          fisher: Boolean = false,
```
nh13, line 128: `fisher` -> `fisherJava` here, or `fisherJava` -> `fisher` below

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -345,7 +349,8 @@ class VarDictJavaEndToEnd
       pileupMode          = pileupMode,
       countNsInTotalDepth = countNsInTotalDepth,
       minThreads          = minThreads,
-      maxThreads          = maxThreads
+      maxThreads          = maxThreads,
+      fisher              = fisherJava
```
nh13: make the same, don't care which way

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -365,7 +370,7 @@ class VarDictJavaEndToEnd
           minimumAf                         = minimumAf,
           printEndTag                       = false
         )
-        (bias | var2VcfValid)
+        if (fisherJava) var2VcfValid else (bias | var2VcfValid)
```
nh13: Move `bias` to only be created if it needs to be.
```scala
if (fisher) var2VcfValid else {
  val bias         = new ShellCommand(VarDictJava.TestStrandBias.toString) with PipeWithNoResources[Text, Text]
  (bias | var2VcfValid)
}
```

tasks/src/main/scala/dagr/tasks/vc/VarDictJava.scala
```diff
@@ -381,7 +386,7 @@ class VarDictJavaEndToEnd
           minimumHighQualityAltDepth        = minimumHighQualityAltDepth,
           minimumAf                         = minimumAf
         )
-        (somatic | var2VcfPaired)
+        if (fisherJava) var2VcfPaired else (somatic | var2VcfPaired)
```
nh13: Ditto about `somatic`

nh13 (changes requested):
Minor changes, back to you @mjhipp

mjhipp:
Thanks @nh13, should be all fixed up now.

PR #379, opened 2020-07-02
Bugfix/nested non unit tasks

nh13:
@tfenne a very interesting edge case I hit in the real world

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
@@ -290,6 +290,7 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
   private[execsystem] def processCompletedTask(taskId: TaskId, doRetry: Boolean = true): Unit = {
     val node      = this(taskId)
     val taskInfo  = node.taskInfo
+    logger.debug("processCompletedTask")
     logTaskMessage(taskInfo=taskInfo)
     val updateNodeToCompleted: Boolean = if (TaskStatus.isTaskFailed(taskStatus = taskInfo.status) && doRetry) {
       val retryTask = taskInfo.task match {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -290,6 +290,7 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
   private[execsystem] def processCompletedTask(taskId: TaskId, doRetry: Boolean = true): Unit = {
     val node      = this(taskId)
     val taskInfo  = node.taskInfo
+    logger.debug("processCompletedTask")
```
tfenne: Wouldn't it be more useful to also log the taskId?

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -327,20 +328,21 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
     * @return for each completed task, a map from a task identifier to a tuple of the exit code and the status of the `onComplete` method
     */
   private def updateCompletedTasks(): Map[TaskId, (Int, Boolean)] = {
-    val completedTasks: Map[TaskId, (Int, Boolean)] = taskExecutionRunner.completedTasks()
-    val emptyTasks = graphNodesInStateFor(GraphNodeState.NO_PREDECESSORS).filter(_.task.isInstanceOf[Task.EmptyTask])
-
-    emptyTasks.foreach { node =>
+    // Get the tasks that are not eligible to be scheduled/executed, and have no predecessors (can be completed).
+    val toCompleteTasks = graphNodesInStateFor(GraphNodeState.NO_PREDECESSORS).filterNot(_.task.isInstanceOf[UnitTask])
```
tfenne: This is confusing to me ... wouldn't this also grab e.g. `Pipeline` instances that have no predecessors and complete them?  If not, can you add to the comment as to how this doesn't select those?

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -327,20 +328,21 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
...
+    toCompleteTasks.foreach { node =>
       node.taskInfo.status = TaskStatus.SUCCEEDED
       processCompletedTask(node.taskId)
-      logger.debug("updateCompletedTasks: empty task [" + node.task.name + "] completed")
+      logger.debug("updateCompletedTasks: to-complete task [" + node.task.name + "] completed")
```
tfenne: Maybe just drop `empty`/`to-complete`?

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -327,20 +328,21 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
...
     }
 
+    // Get the tasks that have been executed to completion.
+    val completedTasks: Map[TaskId, (Int, Boolean)] = taskExecutionRunner.completedTasks()
```
tfenne: Why not just have a single loop with:
 `(toCompleteTasks.map(_.task_id) ++ completedTasks.keys).foreach { taskId => ... }`
I get that the log messages are different, but you could just make them the same and I think it would be clearer that you're not actually doing different handling.

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -562,17 +564,22 @@ class TaskManager(taskManagerResources: SystemResources = TaskManagerDefaults.de
     // get the running tasks to estimate currently used resources
     val runningTasks: Map[UnitTask, ResourceSet] = runningTasksMap
 
-    // get the tasks that are eligible for execution (tasks with no dependents)
-    val (emptyTasks: Seq[Task], readyTasks: Seq[Task]) = {
-      graphNodesInStateFor(NO_PREDECESSORS).map(_.task).toSeq.partition(_.isInstanceOf[Task.EmptyTask])
+    // get the tasks that are eligible for execution (tasks with no dependents).  UnitTasks should be scheduled for
+    // execution, while other tasks (such as EmptyTask) should be treated as though they are "RUNNING".
+    val (toScheduleTasks: Seq[UnitTask], readyToComplete: Seq[UnitTask]) = {
+      graphNodesInStateFor(NO_PREDECESSORS).map(_.task).toSeq.partition {
```
tfenne: Could just do `_.isInstaneceOf[UnitTask]`

tfenne (approved):
A bunch of comments.  Take what you will, leave the rest and then merge it!

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
nh13: Removed

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
nh13: Yes, I'll put in a developer note while I have the reasoning fresh in my brain.  The short of it is when we havea `Pipeline`-like task ZZZ (tasks that return other tasks via `getTasks), the returned tasks via `getTasks` get added as _predecessors of `ZZZ`, such that only when the child tasks are complete will `ZZZ` be in `NO_PREDECESSORS` state and thus returned here.  The reason we didn't have an issue previously is because we didn't have a situation where we only had "empty" tasks read to run.

PR #380, opened 2020-07-07
Add flag to allow for the output of all alignments in bwa mem. 

jacarey:
Additional alignments are output as secondary alignments.


tfenne (approved):
Could you move the parameter above the thread and memory ones though please?

PR #385, opened 2021-01-24
Always use the logical processor count for system max cores

### LINKED ISSUE #370: Why does Dagr use the physical processor count instead of logical for Macs only?
On my MacBook Pro with hyper-threading enabled I have 8 physical cores yet 16 addressable cores.

```
❯ sysctl -n hw.physicalcpu
8
❯ sysctl -n hw.logicalcpu
16
```

https://github.com/fulcrumgenomics/dagr/blob/e2a634a5054d4a2200aa360ea33843280b2b0b51/core/src/main/scala/dagr/core/execsystem/Resource.scala#L33-L38

I may be under-utilizing my local compute resources if capped to 8 cores instead of the full 16. Here's a screenshot of when I ask Dagr to run a task with the maximum allowed system cores:

Screen Shot 2020-01-05 at 5 21 24 PM

Since hyper-threading is not unique to Macs, I am wondering why this design decision was made.

clintval:
I have an 8-core MacOS with hyperthreading (16 virtual cores total).

I'd like to tell my tasks to use them all!

Running `ExamplePipeline` before this PR:

```
[2021/01/23 22:40:21 | TaskManager | Info] Executing with 8.0 cores and 24g system memory.
```

And after:

```
[2021/01/23 22:45:31 | TaskManager | Info] Executing with 16.0 cores and 24g system memory.
```

Closes #370.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@
 
 # Build tool specific
 target
+
+# jEnv
+.java-version
--- a/core/src/main/scala/dagr/core/execsystem/Resource.scala
+++ b/core/src/main/scala/dagr/core/execsystem/Resource.scala
@@ -24,18 +24,13 @@
 package dagr.core.execsystem
 
 import oshi.SystemInfo
-import oshi.hardware.platform.mac.MacHardwareAbstractionLayer
 
 /** Manipulates system resources */
 object Resource {
   private val hal = new SystemInfo().getHardware
 
   /** Total number of cores in the system */
-  val systemCores : Cores =
-    if (hal.isInstanceOf[MacHardwareAbstractionLayer])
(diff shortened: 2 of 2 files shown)


nh13:
@clintval thanks for this PR, let us think about if we ant cores to be the # of physical cores, or the # of virtual cores/threads.

tfenne:
I'm 👍 .  I think I was the one who wanted physical cores originally, and back in the earlier days of hyper-threading it wasn't really clear to me how useful it was.  However, nowadays my habit it to, e.g. run `bwa` on my mac with 12 threads, not 6.  So I realized I'm doing this when running things manually.

If we wanted to we could always have a config setting for `dagr.core_type=[logical|physical]` but I don't think that's really necessary.

PR #386, opened 2021-01-24
Round cores in log statements to 2 decimals of precision

### LINKED ISSUE #323: TaskManager states cores used as a fraction with floating point error
I have a task which requires 1 core but when executed on a 40 core machine I get the following log:
```
[2018/04/03 16:46:28 | TaskManager | Info] 'FastqToUnmappedBam.1.1' has been started on attempt #1 with 4.300000000000001 cores and 8g memory
```
How many cores are actually used in this scenario? Shouldn't the number at least be an integer? Let me know if you need more contextual information. It would be hard to provide a code sample since this task is embedded in a larger workflow.

clintval:
Spring cleaning motivated this PR. Closes #323.

### DIFF
--- a/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
+++ b/core/src/main/scala/dagr/core/execsystem/TaskManager.scala
@@ -71,9 +71,11 @@ case class SystemResources(cores: Cores, systemMemory: Memory, jvmMemory: Memory
 object TaskManagerDefaults extends LazyLogging {
   def defaultTaskManagerResources: SystemResources = {
     val resources = SystemResources(cores=None, totalMemory=None) // Let the apply method figure it all out
-    val cores     = TaskManager.roundCoresForLogging(resources.cores)
-    logger.info("Defaulting System Resources to " + cores.toString + " cores and " + Resource.parseBytesToSize(resources.systemMemory.value) + " memory")
-    logger.info("Defaulting JVM Resources to " + Resource.parseBytesToSize(resources.jvmMemory.value) + " memory")
+    val cores     = resources.cores.value
+    val sysMemory = Resource.parseBytesToSize(resources.systemMemory.value)
+    val jvmMemory = Resource.parseBytesToSize(resources.jvmMemory.value)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dagr/core/execsystem/TaskManager.scala
```diff
@@ -70,7 +71,8 @@ case class SystemResources(cores: Cores, systemMemory: Memory, jvmMemory: Memory
 object TaskManagerDefaults extends LazyLogging {
   def defaultTaskManagerResources: SystemResources = {
     val resources = SystemResources(cores=None, totalMemory=None) // Let the apply method figure it all out
-    logger.info("Defaulting System Resources to " + resources.cores.value + " cores and " + Resource.parseBytesToSize(resources.systemMemory.value) + " memory")
+    val cores     = TaskManager.roundCoresForLogging(resources.cores)
```
nh13: why not just `f"${resources.cores}%.2f"`  here and everywhere?

nh13 (changes requested):

core/src/main/scala/dagr/core/execsystem/TaskManager.scala
clintval: Could do! If I used `printf`-style formatting then you would **always** have 2 decimals, even if you didn't need them. For example, with `printf`-style formatting, if you had 4 cores exactly, you would log that as `4.00` whereas with the current implementation, it would be simply `4.0`.

clintval: Up to you, would you rather see trailing zeros?

clintval: Actually, I realized I'm using Doubles which print with at least one trailing zero. So using string formatting like you've suggested would most likely be the cleanest approach all around.

nh13: I didn’t think of that, so a happy accident all around!

clintval: I went with printf-style formatting, 2 decimals places.

PR #398, opened 2021-09-07
Allow callers of Var2VcfValid to turn on printing of a variant's END tag

clintval:
Do you know why historically this was hardcoded to `false`?


nh13 (approved):
I do not know why it was set to false previously.  @tfenne do you recall?  VarDictJava has evolved quite well over the years, so it could have been an issue there, or the fact we were looking only at small variants?

clintval:
To follow up. I have non-exhaustively tested VarDictJava's ability to correctly calculate the `END` tag on a large collection of simple variants (SNV, MNV, insertion, deletion) and on simple symbolics (INS, INV, DUP). Given [that downstream tooling](https://github.com/fulcrumgenomics/fgbio/blob/468a843d19da4f67106dc148a1dba1f2c3390eab/src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala#L104-L105) wants the `END` tag for symbolics, it is a good idea to include it:

> • END : end position of the variant described in this record (for use with symbolic alleles)

Source: https://samtools.github.io/hts-specs/VCFv4.2.pdf

Thank you for the merge!

